PR #11, opened 2015-08-31
Support Filtering of Incoming and Outgoing metrics

udoprog:
This patch implements filtering through a JSON-based filtering language.

This allows the operator to implement filters that targets very specific things, like a specific tag combination.

The `filter` option will not be available in the `input.filter` and `output.filter` parts of the configuration, this is expected to be an array (as exemplified in the changes to `agent/ffwd.yaml`).

### DIFF
--- a/agent/ffwd.yaml
+++ b/agent/ffwd.yaml
@@ -31,6 +31,11 @@ input:
         type: udp
 
 output:
+  filter:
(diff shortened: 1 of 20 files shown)


mehrdad-hassanabadi:
:+1:

PR #14, opened 2016-02-22
[riemann] allow for egress filtering in output plugins

udoprog:
This moves the filter implementations into `ffwd-module` in order to make use of them in plugins.

It also introduces support for injecting a `Filter` into `FlushingPluginSink` and `ProtocolPluginSink` which will be applied if present.

Finally, the `type` filter is introduced, allowing us to filter out specifically events or metrics through `["type", "event"]` or `["type", "metric"]`.


parmus:
+1

PR #20, opened 2016-06-29
Parmus/refactor tags

parmus:
@udoprog


udoprog:
Variable names should be `camelCase`.

Issue #18, opened 2016-05-23
add findbugs and checkstyle to build


udoprog:

dmichel1:
Findbugs is being added as part of https://github.com/spotify/ffwd/pull/99

dmichel1:
Checkstyle is being included from the [foss-root](https://github.com/spotify/foss-root/blob/master/pom.xml#L127)

PR #21, opened 2016-06-29
Introduce more build checks

udoprog:
This solves #18

Regarding the ffwd-module -> ffwd-api, this is the convention we have in other projects that has an API. And `module` is a generic keyword in Maven which would make it confusing.


parmus:
+1

PR #29, opened 2016-12-15
Distribute nodes across Kafka partitions more evenly

zfrank:
The current mechanism to assign metrics to a Kafka partition
based on the hostname.hashCode() does not generate a distribution even
enough and create some hot spots in our Kafka cluster. By calculating a
CRC32 of the hostname and using that instead we create a more even
distribution, according to our tests.

### DIFF
--- a/modules/kafka/src/main/java/com/spotify/ffwd/kafka/KafkaPartitioner.java
+++ b/modules/kafka/src/main/java/com/spotify/ffwd/kafka/KafkaPartitioner.java
@@ -24,6 +24,7 @@ import com.spotify.ffwd.model.Metric;
 
 import java.util.Optional;
 import java.util.function.Supplier;
+import java.util.zip.CRC32;
 
 @JsonTypeInfo(use = JsonTypeInfo.Id.NAME, include = JsonTypeInfo.As.PROPERTY, property = "type")
 @JsonSubTypes({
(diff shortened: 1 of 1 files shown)


parmus (approved):
LGTM

modules/kafka/src/main/java/com/spotify/ffwd/kafka/KafkaPartitioner.java
```diff
@@ -37,13 +38,18 @@
     int partition(final Metric metric);
 
     class Host implements KafkaPartitioner {
+        CRC32 crcGenerator;
```
udoprog: Since this doesn't depend on anything in the constructor, it can be:
```java
private final CRC32 crcGenerator = new CRC32();
```

udoprog (approved):
Nitpick, but still approved.

modules/kafka/src/main/java/com/spotify/ffwd/kafka/KafkaPartitioner.java
```diff
@@ -37,13 +38,18 @@
...
     class Host implements KafkaPartitioner {
+        private final CRC32 crcGenerator;
+
         @JsonCreator
         public Host() {
+            crcGenerator = new CRC32();
         }
 
         @Override
         public int partition(final Event event) {
```
udoprog, line 46: Since this is called by multiple threads (I checked now), we can't use a single instance of CRC32, or each thread will race with each other to update/getValue(), possibly causing inconsistent values.

This should have:
```java
final CRC32 crcGenerator = new CRC32();
```

In the method.

udoprog (changes requested):

udoprog (approved):

udoprog:
@zfrank Sorry for the late approval. For some reason I'm not receiving notifications for updates here.

PR #30, opened 2017-01-23
Fix example ffwd.yaml

tchagnon:


tchagnon:
@parmus I was able to run this locally to confirm my metrics, but had to fix the example yaml.

parmus (approved):

PR #40, opened 2017-10-23
[all] introduce new module `http` and Batch support

udoprog:
This introduces a simple `http` module.

The protocol is documented in the README for `modules/http`.


gabrielgerhardsson (approved):
LGTM.
Disclaimer: I was a part of writing this, so would be good if someone else looked at it also.

api/src/main/java/com/spotify/ffwd/output/BatchedPluginSink.java
```diff
@@ -37,4 +38,12 @@
...
      */
     public AsyncFuture<Void> sendMetrics(Collection<Metric> metrics);
+
+    /**
+     * Send the given collection of batches.
+     *
+     * @param batches Batches to send.
+     * @return A future that will be resolved when the batches have been sent.
+     */
+    public AsyncFuture<Void> sendBatches(Collection<Batch> batches);
```
nezdolik, line 48: public keyword is redundant in interface method

api/src/main/java/com/spotify/ffwd/output/FlushingPluginSink.java
```diff
@@ -402,17 +399,23 @@ Batch newBatch() {
     static class Batch {
```
nezdolik, line 399: is it a good name and location for a class that represents collection of batches?

api/src/main/java/com/spotify/ffwd/output/BatchedPluginSink.java
udoprog: Yes, but it follows the convention already established in the file.

api/src/main/java/com/spotify/ffwd/output/FlushingPluginSink.java
udoprog: It's purely internal to this class, so name doesn't matter (much).

PR #43, opened 2017-10-30
More changes to support shim

udoprog:
* Introduces Ribbon for client-side load balancing
* Add mechanism for specifying search domain through configuration.
* Minor bug fixes and improved error messages.

### DIFF
--- a/modules/http/src/main/java/com/spotify/ffwd/http/HttpClient.java
+++ b/modules/http/src/main/java/com/spotify/ffwd/http/HttpClient.java
@@ -1,0 +1,88 @@
+/*
+ * Copyright 2013-2017 Spotify AB. All rights reserved.
+ *
+ * The contents of this file are licensed under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.spotify.ffwd.http;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.spotify.ffwd.model.Batch;
+import eu.toolchain.async.AsyncFramework;
+import eu.toolchain.async.AsyncFuture;
+import eu.toolchain.async.ResolvableFuture;
+import java.io.IOException;
+import lombok.Data;
+import okhttp3.Call;
+import okhttp3.Callback;
+import okhttp3.MediaType;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+
+@Data
+public class HttpClient {
+    public static final String V1_BATCH_ENDPOINT = "v1/batch";
+    public static final String PING_ENDPOINT = "ping";
+
+    private final AsyncFramework async;
+    private final ObjectMapper mapper;
+    private final OkHttpClient httpClient;
+    private final String baseUrl;
+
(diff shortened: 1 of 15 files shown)


modules/http/src/main/java/com/spotify/ffwd/http/HttpClient.java
```diff
@@ -0,0 +1,88 @@
...
+import okhttp3.MediaType;
+import okhttp3.OkHttpClient;
+import okhttp3.Request;
+import okhttp3.RequestBody;
+import okhttp3.Response;
+
+@Data
+public class HttpClient {
+    public static final String V1_BATCH_ENDPOINT = "v1/batch";
+    public static final String PING = "ping";
```
nezdolik: for consistent naming, better name it PING_ENDPOINT

gabrielgerhardsson (approved):

modules/http/src/main/java/com/spotify/ffwd/http/HttpClient.java
```diff
@@ -0,0 +1,88 @@
...
+    public AsyncFuture<Void> ping() {
+        final Request.Builder request = new Request.Builder();
+
+        request.url(baseUrl + "/" + PING);
+        request.get();
+
+        return execute(request);
+    }
+
+    private AsyncFuture<Void> execute(final Request.Builder request) {
```
nezdolik, line 70: why not to pass builded Request object here?

modules/http/src/main/java/com/spotify/ffwd/http/HttpDecoder.java
```diff
@@ -87,4 +95,13 @@ private void postBatch(
             .writeAndFlush(new DefaultFullHttpResponse(HttpVersion.HTTP_1_1, HttpResponseStatus.OK))
             .addListener((ChannelFutureListener) future -> future.channel().close());
     }
+
+    private void getPing(
```
nezdolik, line 99: params in and out are not used

modules/http/src/main/java/com/spotify/ffwd/http/HttpPing.java
```diff
@@ -0,0 +1,40 @@
...
+public class HttpPing implements IPing {
+    @Inject
+    HttpClientFactory clientFactory;
+
+    @Override
+    public boolean isAlive(final Server server) {
+        try {
+            clientFactory.newClient(server).ping().get();
+        } catch (Exception e) {
+            log.warn("Error when pinging server ({}): {}", server, e.getMessage());
```
nezdolik, line 33: logging same msg at 2 diff log levels

modules/http/src/main/java/com/spotify/ffwd/http/HttpClient.java
udoprog: Probably in case some kind of consistent modification should be made prior to sending it. Like adding a header.

modules/http/src/main/java/com/spotify/ffwd/http/HttpDecoder.java
udoprog: No, but it makes the methods handling requests consistent.

modules/http/src/main/java/com/spotify/ffwd/http/HttpPing.java
udoprog: They are subtly different. One will only include the message of the error (`warn`), the other will include the stacktrace (`trace`). This exception is expected to happen fairly often, so logging the stacktrace by default would be spammy.

udoprog: Kind of a hacky solution, but it works.

modules/http/src/main/java/com/spotify/ffwd/http/HttpClient.java
udoprog: Yep, changed. Thanks.

nezdolik (approved):

modules/http/src/main/java/com/spotify/ffwd/http/HttpClient.java
nezdolik: then the name 'execute' is misleading :)

PR #51, opened 2017-12-14
 [http] close response and propagate errors in modules/http

udoprog:
This (hopefully) fixes an issue reported by @mattnworb:
```
WARNING: A connection to <url> was leaked. Did you forget to close a response body? To see where this was allocated, set the OkHttpClient logger level to FINE: Logger.getLogger(OkHttpClient.class.getName()).setLevel(Level.FINE);
```
Note: The client code in this module will soon be replaced with [ffwd-http-client](http://search.maven.org/#search%7Cgav%7C1%7Cg%3A%22com.spotify.ffwd%22%20AND%20a%3A%22ffwd-http-client%22) which is a cleanup and extraction of the code in the HTTP module that can also be used by third party projects.

### DIFF
--- a/ffwd-http-client/src/main/java/com/spotify/ffwd/http/RawHttpClient.java
+++ b/ffwd-http-client/src/main/java/com/spotify/ffwd/http/RawHttpClient.java
@@ -85,6 +85,8 @@ public class RawHttpClient {
                                 "HTTP request failed: " + response.code() + ": " +
                                     response.message()));
                         }
+
+                        response.close();
                     }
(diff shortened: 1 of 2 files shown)


mattnworb:
👍  seems to match up with what [the okhttp3 docs say](https://github.com/square/okhttp/blob/17485bc41852ee9586ebe167555737825a0d1043/okhttp/src/main/java/okhttp3/Callback.java#L30-L31)

nezdolik (approved):

PR #61, opened 2018-01-24
[all] Introduce 'batching' conf for outputs

gabrielgerhardsson:
Instead of specifying 'flushInterval' directly on an output plugin,
it's now possible to optionally move it within a sub structure
'batching':

```
output:
  plugins:
    - type: debug
      batching:
        flushInterval: 10000
```

This will allow adding further related configuration options here.

The old way of specifying flushInterval outside of batching is still
supported. The two methods are mutually exclusive tho.

### DIFF
--- a/api/src/main/java/com/spotify/ffwd/module/Batching.java
+++ b/api/src/main/java/com/spotify/ffwd/module/Batching.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2018 Spotify AB. All rights reserved.
+ *
+ * The contents of this file are licensed under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.spotify.ffwd.module;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import java.util.Optional;
+import lombok.Data;
+
+@Data
+public class Batching {
+    protected final Optional<Long> flushInterval;
+    protected final Optional<Long> batchSizeLimit;
+    protected final Optional<Long> maxPendingFlushes;
+
+    @JsonCreator
+    public Batching(
+        @JsonProperty("flushInterval") Optional<Long> flushInterval,
+        @JsonProperty("batchSizeLimit") Optional<Long> batchSizeLimit,
+        @JsonProperty("maxPendingFlushes") Optional<Long> maxPendingFlushes
+    ) {
+        this.flushInterval = flushInterval;
+        this.batchSizeLimit = batchSizeLimit;
+        this.maxPendingFlushes = maxPendingFlushes;
+    }
+
+    public static Batching empty() {
+        return from(Optional.empty(), Optional.empty(), Optional.empty());
+    }
+
+    public static Batching from(
+        final Optional<Long> flushInterval, final Optional<Batching> batching
+    ) {
+        return from(flushInterval, batching, Optional.empty());
+    }
+
+    public static Batching from(
+        final Optional<Long> flushInterval, final Optional<Batching> batching,
+        final Optional<Long> defaultFlushInterval
+    ) {
+        if (flushInterval.isPresent() && batching.isPresent()) {
+            throw new RuntimeException(
+                "Can't have both 'batching' and 'flushInterval' on the same level in the " +
+                    "configuration. Maybe move 'flushInterval' into 'batching'?");
+        }
+        if (batching.isPresent()) {
+            return batching.get();
+        }
+        if (flushInterval.isPresent()) {
+            return new Batching(flushInterval, Optional.empty(), Optional.empty());
+        }
+        return new Batching(defaultFlushInterval, Optional.empty(), Optional.empty());
+    }
+}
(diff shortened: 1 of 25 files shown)


api/src/main/java/com/spotify/ffwd/output/OutputPlugin.java
```diff
@@ -23,24 +23,31 @@
...
     }
 
     public OutputPlugin(
-        final Optional<Filter> filter, final Optional<Long> flushInterval
+        final Optional<Filter> filter, final Batching batching
     ) {
         this.filter = filter;
-        this.flushInterval = flushInterval;
+        this.batching = batching;
+        if (filter == null) {
```
nezdolik: filter could be optional in configuration

api/src/main/java/com/spotify/ffwd/output/OutputPlugin.java
```diff
@@ -62,35 +69,38 @@ public OutputPlugin(
...
                         .to(batchedPluginSink);
-                    bind(flushingKey).toInstance(new FlushingPluginSink(flushInterval.get()));
+                    bind(flushingKey).toInstance(
+                        new BatchingPluginSink(batching.getFlushInterval().get(),
+                            batching.getBatchSizeLimit(), batching.getMaxPendingFlushes()));
 
                     sinkKey = flushingKey;
                 }
 
-                if (filter != null && filter.isPresent()) {
```
nezdolik, line 93: filter can be empty

core/src/main/java/com/spotify/ffwd/debug/DebugOutputPlugin.java
```diff
@@ -48,7 +50,7 @@ public Module module(final Key<PluginSink> key, final String id) {
             protected void configure() {
                 final Key<DebugPluginSink> sinkKey =
                     Key.get(DebugPluginSink.class, Names.named("debugSink"));
-                bind(sinkKey).to(DebugPluginSink.class);
+                bind(sinkKey).to(DebugPluginSink.class).in(Scopes.SINGLETON);
```
nezdolik, line 53: this limits possibility to declare multiple debug output plugins

modules/http/src/main/java/com/spotify/ffwd/http/HttpOutputPlugin.java
```diff
@@ -79,7 +81,7 @@ public ILoadBalancer setupRibbonClient(
             protected void configure() {
                 final Key<HttpPluginSink> sinkKey =
                     Key.get(HttpPluginSink.class, Names.named("httpSink"));
-                bind(sinkKey).to(HttpPluginSink.class);
+                bind(sinkKey).to(HttpPluginSink.class).in(Scopes.SINGLETON);
```
nezdolik, line 84: this limits possibility to declare multiple http output plugins

modules/signalfx/src/main/java/com/spotify/ffwd/signalfx/SignalFxOutputPlugin.java
```diff
@@ -103,7 +105,7 @@ public Module module(final Key<PluginSink> key, final String id) {
             protected void configure() {
                 final Key<SignalFxPluginSink> sinkKey =
                     Key.get(SignalFxPluginSink.class, Names.named("signalfxSink"));
-                bind(sinkKey).to(SignalFxPluginSink.class);
+                bind(sinkKey).to(SignalFxPluginSink.class).in(Scopes.SINGLETON);
```
nezdolik, line 108: this limits possibility to declare multiple signalfx output plugins

api/src/main/java/com/spotify/ffwd/module/Batching.java
```diff
@@ -0,0 +1,67 @@
...
+        return from(Optional.empty(), Optional.empty(), Optional.empty());
+    }
+
+    public static Batching from(
+        final Optional<Long> flushInterval, final Optional<Batching> batching
+    ) {
+        return from(flushInterval, batching, Optional.empty());
+    }
+
+    public static Batching from(
```
nezdolik: maybe add some javadoc to method?

nezdolik (changes requested):

api/src/main/java/com/spotify/ffwd/module/Batching.java
gabrielgerhardsson: Thanks, fixed.

api/src/main/java/com/spotify/ffwd/output/OutputPlugin.java
gabrielgerhardsson: I don't think so. It's an optional, so it will be Optional.empty() if it wasn't specified in the configuration. It will not be null.
I actually remember thinking previously that this was unnecessary, so thank you, I was reminded now :) I've removed this check now.

api/src/main/java/com/spotify/ffwd/output/OutputPlugin.java
gabrielgerhardsson: I don't think so. It's an optional, so it will be Optional.empty() if it wasn't specified in the configuration. It will not be null.

core/src/main/java/com/spotify/ffwd/debug/DebugOutputPlugin.java
gabrielgerhardsson: This was a deliberate decision on the input of @udoprog. I believe that the reasoning is that there's no need to have multiple such output plugins specified.

modules/http/src/main/java/com/spotify/ffwd/http/HttpOutputPlugin.java
gabrielgerhardsson: This was a deliberate decision on the input of @udoprog. I believe that the reasoning is that there's no need to have multiple such output plugins specified.

modules/signalfx/src/main/java/com/spotify/ffwd/signalfx/SignalFxOutputPlugin.java
gabrielgerhardsson: This was a deliberate decision on the input of @udoprog. I believe that the reasoning is that there's no need to have multiple such output plugins specified.

nezdolik (approved):

PR #62, opened 2018-01-30
[agent,api] Add optional metrics for batching

gabrielgerhardsson:
Every output plugin can have an optional BatchingPluginSink in front
of it. This changes the behavior of the output, and could under high
load be a place where data is built up in queues and outstanding
writes - so it's interesting to monitor.
This commit adds some new disabled-by-default metrics for this. Enable
by specifying reportStatistics:true within the batching block for the
relevant output. Like so:

```
output:
  plugins:
    - type: debug
      batching:
        flushInterval: 10000
        reportStatistics: true
```

### DIFF
--- a/api/src/main/java/com/spotify/ffwd/output/BatchingPluginSink.java
+++ b/api/src/main/java/com/spotify/ffwd/output/BatchingPluginSink.java
@@ -20,6 +20,7 @@ import com.google.inject.name.Named;
 import com.spotify.ffwd.filter.Filter;
 import com.spotify.ffwd.model.Event;
 import com.spotify.ffwd.model.Metric;
+import com.spotify.ffwd.statistics.BatchingStatistics;
 import com.spotify.ffwd.statistics.OutputPluginStatistics;
 import eu.toolchain.async.AsyncFramework;
 import eu.toolchain.async.AsyncFuture;
@@ -69,6 +70,9 @@ public class BatchingPluginSink implements PluginSink {
 
     @Inject
     OutputPluginStatistics statistics;
+
+    @Inject
+    BatchingStatistics batchingStatistics;
 
     @Named("flushing")
     @Inject(optional = true)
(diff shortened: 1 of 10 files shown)


gabrielgerhardsson:
The main complicating thing with this solution is that I wanted the new metrics for batching to be disabled by default, while the implementation was supplied with DI. I ended up injecting a supplier. That way I could get the arguments I needed in the injection scope (OutputPluginModule), but only create the object (and thus register the new metrics in the registry) when it actually is wanted.

api/src/main/java/com/spotify/ffwd/output/BatchingPluginSink.java
```diff
@@ -70,6 +74,11 @@
     @Inject
     OutputPluginStatistics statistics;
 
+    @Inject
+    Supplier<BatchingStatistics> batchingStatisticsSupplier;
+
+    BatchingStatistics batchingStatistics = new NoopCoreStatistics().newBatching("");
```
udoprog: Could we instead provide this through the injected supplier? I.e. if `reportStatistics` is `false`, provide a `Supplier` that returns a noop impl?

That way, we don't have to pass `reportStatistics` into this class, making it a bit more encapsulated.

udoprog (approved):

udoprog:
Looks really good. Thanks for this!

api/src/main/java/com/spotify/ffwd/output/BatchingPluginSink.java
gabrielgerhardsson: Thanks for the feedback! I've given it a try. Please see the separate commit :)

api/src/main/java/com/spotify/ffwd/output/OutputPlugin.java
```diff
@@ -104,6 +110,20 @@ protected void configure() {
...
+            @Singleton
+            public BatchingStatistics batchingStatisticsProvider(
+                CoreStatistics statistics, @Named("pluginId") String pluginId, Logger log
+            ) {
+                if (batching.isReportStatistics()) {
+                    log.info("Enabling reporting of batching-specific metrics for output " +
+                        input.getTypeLiteral().getRawType().getName());
+                    return statistics.newBatching(pluginId);
+                } else {
+                    return new NoopCoreStatistics().newBatching(pluginId);
```
udoprog: This is really only a nit, but the NOOP batching singleton could be exposed here so that the pluginId does not have to be provided.

gabrielgerhardsson: Gotcha :) A small change so I did it.

PR #65, opened 2018-02-09
[batching] Use write-batch-size metric

gabrielgerhardsson:


api/src/main/java/com/spotify/ffwd/output/BatchingPluginSink.java
```diff
@@ -396,10 +396,10 @@ public void run() {
         }
 
         // chain into batch future.
-        return async
-            .collectAndDiscard(futures)
-            .onFinished(writeMonitor)
-            .onFinished(() -> batchingStatistics.reportQueueSizeDec(batch.size()));
+        return async.collectAndDiscard(futures).onFinished(writeMonitor).onFinished(() -> {
+            batchingStatistics.reportQueueSizeDec(batch.size());
+            batchingStatistics.reportInternalBatchWrite(batch.size());
```
udoprog, line 401: Might be cleaner to model these after intent. Like a, `decreaseBatchSize` that only internally decrements these.

But I'm good with this!

udoprog (approved):

api/src/main/java/com/spotify/ffwd/output/BatchingPluginSink.java
gabrielgerhardsson: Yeah it's starting to get kludgy. I'll merge this tho, and we can look at making it nicer.

PR #66, opened 2018-02-19
[core] Add 'tagsToResource' mapping

gabrielgerhardsson:
tagsToResource configuration makes it possible to selectively map
certain tags as resource identifiers.
As a special case, it's also possible to map the 'host' field with
this configuration, even tho it's not exactly a tag.

### DIFF
--- a/core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
+++ b/core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
@@ -53,6 +53,10 @@ public class CoreOutputManager implements OutputManager {
     private Map<String, String> tags;
 
     @Inject
+    @Named("tagsToResource")
+    private Map<String, String> tagsToResource;
+
+    @Inject
     @Named("resource")
     private Map<String, String> resource;
 
@@ -204,6 +208,8 @@ public class CoreOutputManager implements OutputManager {
         final Map<String, String> mergedResource = Maps.newHashMap(resource);
         mergedResource.putAll(metric.getResource());
 
+        processTagsToResource(mergedTags, mergedResource);
+
         final Set<String> mergedRiemannTags = Sets.newHashSet(riemannTags);
         mergedRiemannTags.addAll(metric.getRiemannTags());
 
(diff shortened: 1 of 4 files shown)


core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -241,7 +249,53 @@ private Batch filter(final Batch batch) {
...
+            // wanted name.
+            if (tags.containsKey(fromTag) && !resource.containsKey(asResource)) {
+                resource.put(asResource, tags.get(fromTag));
+            }
+            // Always remove the tag
+            tags.remove(fromTag);
+        });
+    }
+
+    private String selectHost(final Metric metric, final Map<String, String> resource) {
```
udoprog: Might be time to consider making host just another tag internally.

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -241,7 +249,53 @@ private Batch filter(final Batch batch) {
...
+     * @param resource Map of resource identifiers that will be used when constructing a metric
+     */
+    private void processTagsToResource(
+        final Map<String, String> tags, final Map<String, String> resource
+    ) {
+        if (tagsToResource.isEmpty()) {
+            return;
+        }
+
+        tagsToResource.forEach((fromTag, asResource) -> {
```
udoprog: This can be made simpler using the fact that `remove` returns null if the mapping doesn't exist, and we have `putIfAbsent` at our disposal.

```java
final String tag = tags.remove(fromTag);

if (tag != null) {
  resource.putIfAbsent(toResource, tag);
}
```

udoprog (approved):
Nothing big. Just suggestions.

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
gabrielgerhardsson: Thanks, fixed now.

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
gabrielgerhardsson: True. I'll give it a quick stab to see if it's cumbersome or not.

gabrielgerhardsson: I submitted #68, to make host be a tag. I suggest we finalise that one before merging #66.

nezdolik (commented):
LGTM

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
nezdolik: change not visible in PR

PR #67, opened 2018-02-19
[agent] Histogram supporting low rate scenarios

gabrielgerhardsson:
If the metric isn't emitted so often, a traditional Histogram with
exponential backoff reservoir will leave extreme values lingering
for quite long. This implementation uses a TTL reservoir that only
keeps metrics for a limited time.

### DIFF
--- a/agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
+++ b/agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
@@ -18,7 +18,10 @@ package com.spotify.ffwd.statistics;
 import com.codahale.metrics.Counter;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Meter;
+import com.codahale.metrics.Metric;
+import com.codahale.metrics.SlidingTimeWindowReservoir;
 import com.spotify.metrics.core.MetricId;
+import com.spotify.metrics.core.SemanticMetricBuilder;
 import com.spotify.metrics.core.SemanticMetricRegistry;
 import eu.toolchain.async.FutureFinished;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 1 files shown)


agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
```diff
@@ -18,17 +18,38 @@
...
 
 @RequiredArgsConstructor
 public class SemanticCoreStatistics implements CoreStatistics {
+    private static final int HISTOGRAM_TTL_MINUTES = 2;
     private final SemanticMetricRegistry registry;
     private final MetricId metric = MetricId.build();
 
+    /* A Histogram with a Reservoir that remembers values for a maximum of x minutes, so that slow
+     * rate doesn't cause long-lingering values in the Histogram. */
+    private static final SemanticMetricBuilder<Histogram> HISTOGRAM_BUILDER =
```
xeago: What is the reason for instantiating an anonymous class over having a named class?

xeago (approved):

agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
gabrielgerhardsson: Thanks for your review! 
My reasoning is that it's a very simple class. Ideally, it would have been possible to specify a reservoir to the default registry.histogram() method, but it's not. This is the most simple workaround to that. The upside of having it in the same file as the other related things is in my opinion greater than any added clarity that a separate file would give.

gabrielgerhardsson: I realise I didn't actually answer your question :) Yes, indeed, it could be a named class in the same file. In this case, tho, It would in my opinion just add lines and not clarity. Do you disagree?

PR #68, opened 2018-02-23
[all] Make 'host' a tag for metrics

gabrielgerhardsson:
We used to have a separate field for 'host'. No more. It's now just
a normal tag.

### DIFF
--- a/core/src/test/java/com/spotify/ffwd/output/OutputManagerTest.java
+++ b/core/src/test/java/com/spotify/ffwd/output/OutputManagerTest.java
@@ -1,0 +1,158 @@
+package com.spotify.ffwd.output;
+
+import static org.junit.Assert.assertEquals;
+import static org.mockito.Matchers.any;
+import static org.mockito.Mockito.doNothing;
+import static org.mockito.Mockito.times;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+import com.google.common.collect.ImmutableList;
+import com.google.common.collect.ImmutableMap;
+import com.google.common.collect.ImmutableSet;
+import com.google.common.collect.Lists;
+import com.google.inject.AbstractModule;
+import com.google.inject.Guice;
+import com.google.inject.Injector;
+import com.google.inject.Module;
+import com.google.inject.TypeLiteral;
+import com.google.inject.name.Names;
+import com.spotify.ffwd.debug.DebugServer;
+import com.spotify.ffwd.filter.Filter;
+import com.spotify.ffwd.model.Batch;
+import com.spotify.ffwd.model.Event;
+import com.spotify.ffwd.model.Metric;
+import com.spotify.ffwd.statistics.OutputManagerStatistics;
+import eu.toolchain.async.AsyncFramework;
+import eu.toolchain.async.AsyncFuture;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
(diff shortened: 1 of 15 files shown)


core/src/test/java/com/spotify/ffwd/output/OutputManagerTest.java
```diff
@@ -0,0 +1,187 @@
...
+import org.mockito.runners.MockitoJUnitRunner;
+
+@RunWith(MockitoJUnitRunner.class)
+public class OutputManagerTest {
+    private final static String HOST = "thehost";
+    private final static String KEY = "thekey";
+
+    private List<Metric> sinkMetrics = new ArrayList<>();
+
+    private PluginSink sink = new PluginSink() {
```
nezdolik: would be possible to mock?

modules/signalfx/src/main/java/com/spotify/ffwd/signalfx/SignalFxPluginSink.java
```diff
@@ -112,11 +112,14 @@ public Void call() throws Exception {
                                 .build())
                             .forEach(datapointBuilder::addDimensions);
 
-                        datapointBuilder.addDimensions(SignalFxProtocolBuffers.Dimension
-                            .newBuilder()
-                            .setKey("host")
-                            .setValue(metric.getHost())
-                            .build());
+                        final String host = metric.getTags().get("host");
+                        if (host != null) {
```
nezdolik, line 116: could we ever have host null?

core/src/test/java/com/spotify/ffwd/output/OutputManagerTest.java
gabrielgerhardsson: I tried, but found that it would end up being not so easy to read - something like this:
```
doAnswer(invocation -> {
            final Metric m = ((Metric) invocation.getArguments()[0]);
            sinkMetrics.add(m);
            return (Void)null;
        }).when(sink).sendMetric(any());
```

What do you think?

modules/signalfx/src/main/java/com/spotify/ffwd/signalfx/SignalFxPluginSink.java
gabrielgerhardsson: If there's no host tag at all.

core/src/test/java/com/spotify/ffwd/output/OutputManagerTest.java
nezdolik: You could mock sink and use ArgumentCaptor to verify interactions:
```
    @Test
    public void testSkipTagsForKeys() {
        doNothing().when(sink).sendMetric(any(Metric.class));
        when(sink.isReady()).thenReturn(true);
        ArgumentCaptor<Metric> captor = ArgumentCaptor.forClass(Metric.class);
        skipTagsForKeys = ImmutableSet.of(KEY);
        final OutputManager outputManager = createOutputManager();
        outputManager.sendMetric(m1);
        verify(sink, times(1)).sendMetric(captor.capture());
        Metric sentMetric = captor.getValue();
        assertEquals(
            new Metric(m1.getKey(), m1.getValue(), m1.getTime(), m1.getRiemannTags(), m1.getTags(),
                m1.getResource(), m1.getProc()), sentMetric);
    }
```

See example at [FilteringPluginSinkTest](https://github.com/spotify/ffwd/blob/master/api/src/test/java/com/spotify/ffwd/output/FilteringPluginSinkTest.java)

PR #69, opened 2018-03-12
[all] Really test output sink configuration

gabrielgerhardsson:
These files were missed in an earlier merge. So the only thing merged
in that one was preparatory work to enable testing.

### DIFF
--- a/agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
+++ b/agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
@@ -1,0 +1,135 @@
+/*
+ * Copyright 2013-2017 Spotify AB. All rights reserved.
+ *
+ * The contents of this file are licensed under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with the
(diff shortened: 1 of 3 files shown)


agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
```diff
@@ -0,0 +1,138 @@
...
+
+        final List<List<String>> sinkChains = new ArrayList<>();
+
+        for (final PluginSink sink : sinks) {
+
+            final List<String> sinkChain = new ArrayList<>();
+            sinkChains.add(sinkChain);
+
+            PluginSink sinkToInspect = sink;
+            while (true) {
```
udoprog: Move to a helper function?

udoprog (approved):

agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
gabrielgerhardsson: Thanks for the review @udoprog. I did as you suggested.

PR #70, opened 2018-04-05
[kafka] Switch to kafka-clients 1.0.1

gabrielgerhardsson:
This enables exponential reconnect backoff for the Kafka connection.
Configure with the two Kafka-specific properties reconnect.backoff.ms
and reconnect.backoff.max.ms.

```
output:
  plugins:
    - type: "kafka"
      ...
      producer:
        ...
        reconnect.backoff.ms: 100
        reconnect.backoff.max.ms: 10000
```

The first reconnect will happen reconnect.backoff.ms milliseconds
after the connection fails. Consecutive retries will happen with a
delay that is exponentially increased, up to reconnect.backoff.max.ms.
In each step a 20% random fuzz is added to avoid connection attempt
storms.


modules/kafka/src/main/java/com/spotify/ffwd/kafka/KafkaPluginSink.java
```diff
@@ -76,28 +81,26 @@ public KafkaPluginSink(int batchSize) {
...
 
     @Override
     public void sendEvent(final Event event) {
-        async.call(new Callable<Void>() {
-            @Override
-            public Void call() throws Exception {
-                producer.send(eventConverter.toMessage(event));
-                return null;
-            }
-        }, executorService);
```
nezdolik, line 89: do send on dedicated threadpool as <code>producer.send</code> does await internally and may spend cpu time

modules/kafka/src/main/java/com/spotify/ffwd/kafka/KafkaPluginSink.java
```diff
@@ -76,28 +81,26 @@ public KafkaPluginSink(int batchSize) {
...
     public void sendMetric(final Metric metric) {
-        async.call(new Callable<Void>() {
-            @Override
-            public Void call() throws Exception {
-                producer.send(metricConverter.toMessage(metric));
-                return null;
-            }
-        }, executorService);
+        try {
+            producer.send(metricConverter.toMessage(metric));
```
nezdolik, line 100: do send on dedicated threadpool as <code>producer.send</code> does await internally and may spend cpu time

modules/kafka/src/main/java/com/spotify/ffwd/kafka/KafkaPluginSink.java
```diff
@@ -128,26 +131,50 @@ public void sendBatch(final Batch batch) {
...
+                        future.resolve(null);
+                    } else {
+                        // Wrap to avoid self-suppression when multiple metrics get same exception
+                        future.fail(new Exception(id + ": Failed while sending: ", exception));
+                    }
+                });
+
+                futures.add(future);
             }
-        }, executorService);
```
nezdolik, line 150: do send on dedicated threadpool as <code>producer.send</code> does await internally and may spend cpu time

core/src/main/java/com/spotify/ffwd/AgentCore.java
```diff
@@ -178,8 +178,9 @@ private void start(final Injector primary) throws Exception {
 
         async.collectAndDiscard(startup).get(10, TimeUnit.SECONDS);
 
-        input.init();
+        // Initialize outputs first so that they're ready for getting data
         output.init();
+        input.init();
```
nezdolik, line 183: just feels more logical to init input first, as it passes data to output.

modules/kafka/src/main/java/com/spotify/ffwd/kafka/KafkaPluginSink.java
gabrielgerhardsson: I agree

core/src/main/java/com/spotify/ffwd/AgentCore.java
gabrielgerhardsson: I disagree. An input might start producing data, that is then fed to an output module which hasn't yet been initialised.

dmichel1:
Closing this in favor of opening a new pull request based on the internal commit. I'm working on making this OSS version the source of truth.

@jsferrei @gabrielgerhardsson

PR #74, opened 2018-07-20
 [pubsub] add output plugin

dmichel1:
This plugin allows for outputting metrics to Google Pubsub.

It allows for local testing via the pubsub emulator. See the readme on setup instructions. 

 
Threading appears to be setup correctly as not to block the main ffwd thread (looks okay) but the grpc thread count doesn't seem to be tunable (?). Let's see how this runs on a few instances internally.

screen shot 2018-07-20 at 1 37 41 pm

### DIFF
--- a/modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubOutputPlugin.java
+++ b/modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubOutputPlugin.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright 2013-2018 Spotify AB. All rights reserved.
+ *
+ * The contents of this file are licensed under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with the
+ * License. You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+ * WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+ * License for the specific language governing permissions and limitations under
+ * the License.
+ */
+package com.spotify.ffwd.pubsub;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.google.api.gax.batching.BatchingSettings;
+import com.google.api.gax.core.CredentialsProvider;
+import com.google.api.gax.core.ExecutorProvider;
+import com.google.api.gax.core.FixedCredentialsProvider;
+import com.google.api.gax.core.InstantiatingExecutorProvider;
+import com.google.api.gax.core.NoCredentialsProvider;
+import com.google.api.gax.grpc.GrpcTransportChannel;
+import com.google.api.gax.rpc.FixedTransportChannelProvider;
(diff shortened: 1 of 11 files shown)


modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
```diff
@@ -0,0 +1,142 @@
...
+    return async.resolved();
+  }
+
+  @Override
+  public AsyncFuture<Void> sendBatches(Collection<Batch> batches) {
+    // TODO(dmichel): should this support events?
+    final ArrayList<Metric> metrics = new ArrayList<>();
+
+    batches.forEach(batch -> batch.getPoints().forEach(
+        point -> metrics.add(convertBatchMetric(batch, point))));
```
jsferrei: Since this is a nested `forEach`, I think it would be more readable  formatted as
```
batch -> {
   batch.getPoints().forEach(point -> {
      metrics.add(convertBatchMetric(batch, point));
   }
}
```

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubOutputPlugin.java
```diff
@@ -0,0 +1,147 @@
...
+    return new OutputPluginModule(id) {
+
+      @Provides
+      @Singleton
+      public Publisher publisher() throws IOException {
+        Preconditions.checkArgument(project.isPresent(), "Google project must be set in config");
+        Preconditions.checkArgument(topic.isPresent(), "Pubsub topic must be set in config");
+
+        final Duration publishDelayThreshold = Duration.ofNanos(
+            publishDelayThresholdMs.orElse(DEFAULT_DELAY_THRESHOLD));
```
jsferrei: Perhaps push this `orElse` to the constructor as you did with `this.messageCountBatchSize`

jsferrei (approved):
I think this looks fine, just had a couple of comments

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
dmichel1: fixed formatting 👍

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubOutputPlugin.java
dmichel1: Fixed.

dmichel1:
@jsferrei thanks - I just need to add tests and figure out the async part and then I think this is ready to test on a few instances.

sjoeboo:
So i was just poking around with this and 2 things: 

1). looks like this won't read the env vars to send to the local emulator if set: https://cloud.google.com/pubsub/docs/emulator#accessing_environment_variables

2). since ^^ wasn't working, FFWD was sending metrics basically no where because the test project i had set for the emulator doesn't exist in the "real" PubSub, yet FFWD threw no errors about it not being able to send metrics anywhere.

dmichel1:
@sjoeboo I added emulator support (also in the readme on how to setup) and added a warning on startup if the topic doesn't exist. 

@hexedpackets was working on having the heroic consumers create the topics if they don't exist. The reason the publishes will not is because they will all be running with the least amount of privileges. Also longer term we were thinking about managing the topics and subscriptions in terraform.

sjoeboo:
+1

PR #76, opened 2018-07-31
[kafka] Switch to kafka-clients 1.0.1

dmichel1:
This enables exponential reconnect backoff for the Kafka connection.
Configure with the two Kafka-specific properties reconnect.backoff.ms
and reconnect.backoff.max.ms.

output:
  plugins:
    - type: "kafka"
      ...
      producer:
        ...
        reconnect.backoff.ms: 100
        reconnect.backoff.max.ms: 10000

The first reconnect will happen reconnect.backoff.ms milliseconds
after the connection fails. Consecutive retries will happen with a
delay that is exponentially increased, up to reconnect.backoff.max.ms.
In each step a 20% random fuzz is added to avoid connection attempt
storms.

@jsferrei this was cherry-picked from the commit you made internally.


jsferrei:
+1

PR #80, opened 2018-08-07
[all] bump log4j version

dmichel1:
A newer log4j version is required to delete old log files.

At startup ffwd used to log an error...
`ERROR DefaultRolloverStrategy contains an invalid element or attribute "Delete"`

@jsferrei @sjoeboo


jsferrei:
+1

PR #81, opened 2018-08-09
[ffwd-http-client] split out into its own repo

dmichel1:
This will allow us to version `ffwd` independently from `ffwd-http-client`.

https://github.com/spotify/ffwd-http-client

@jsferrei


jsferrei:
+1

PR #84, opened 2018-08-15
Fixes to the release process

dmichel1:
The main change here is to add the correct release version to the debian package. Previously this was hardcoded to `0.0.1-SNAPSHOT` and would fail with...

```
cp agent/target/ffwd-agent-0.0.1-SNAPSHOT.jar build/ffwd-full.jar
cp: cannot stat ‘agent/target/ffwd-agent-0.0.1-SNAPSHOT.jar’: No such file or directory
```

This also adds instructions on how to perform a release.

@sjoeboo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,5 @@
 .*.swp~
 .*.un~
-target
 *.class
 *.iml
 .classpath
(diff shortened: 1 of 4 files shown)


sjoeboo:
+1

PR #85, opened 2018-08-15
revert kafka-clients 1.0.1

dmichel1:
Internally we have seen issues with the newer Kafka client in regards to batch and fetch settings. Since we are looking to move to Google PubSub ASAP we aren't going to spend time figuring out what changed between client versions.

@sjoeboo @jsferrei


sjoeboo:
+1

PR #87, opened 2018-08-16
Drop "v" from git tag

dmichel1:
The produced jar file that is contained with the debian package is not prefixed with a “v”.

@lmuhlha @Dbochman


Dbochman:
+1

Issue #86, opened 2018-08-16
running ffwd locally needs to use the correct snapshot version


dmichel1:
Currently the version is hardcoded which no longer works because `ffwd` is using semantic versioning.

https://github.com/spotify/ffwd/blob/95bc6c8954d03c63617314908df302729710308a/tools/ffwd#L11

PR #89, opened 2018-09-20
Fix run script so that it always gets the latest version

jsferrei:
Fixes https://github.com/spotify/ffwd/issues/86


dmichel1:
+1

PR #94, opened 2018-10-30
Run travis build on openjdk11

labianchin:


hexedpackets:
Right now ffwd doesn't compile under Java 11. With Java 8 being EOL, we plan to work on making it compatible soon.

PR #97, opened 2018-11-21
ignore events with pubsub; add protobuf serializer

dmichel1:
Two commits:
1. Don't publish events to pubsub since we don't support them
2. Add protobuf metric serializer! Its more cpu and space efficient. It reduces the serialized size of a metric by 18%.

@jsferrei @lmuhlha @hexedpackets


lmuhlha:
👍

PR #99, opened 2019-01-02
Add batching/compression when using pubsub output

dmichel1:
<odify the proto schema to send batches and compresses metrics using the snappy library.  In my testing this reduced the payload to Google Pubsub from 10,976 bytes to 1,984 bytes which is a ~81% savings. This is a breaking change.

@hexedpackets @jsferrei @lmuhlha

### DIFF
--- a/agent/ffwd-local-pubsub-debug.yaml
+++ b/agent/ffwd-local-pubsub-debug.yaml
@@ -6,18 +6,14 @@ input:
     - type: json
       protocol:
         type: udp
+        receiveBufferSize: 26214400
     - type: protobuf
       protocol:
         type: udp
+        receiveBufferSize: 26214400
     - type: protobuf
       protocol:
         type: tcp
-    - type: riemann
-      protocol:
-        type: tcp
-    - type: riemann
-      protocol:
-        type: udp
     - type: http
       protocol:
         type: tcp
@@ -27,8 +23,15 @@ output:
     - type: debug
     - type: pubsub
       serializer:
-        type: spotify100
-      flushInterval: 10000
+        type: spotify100proto
+
+      batching:
+        flushInterval: 10000
+        batchSizeLimit: 10000  # Default: 10000
+        maxPendingFlushes: 10 # Default: 10
+
       project: google-test-project
       topic: test-topic
-      #serviceAccount: /path/to/creds.json
+      requestBytesThreshold: 25000 # Default: 5000
+      messageCountBatchSize: 10 # Default: 1000
+      publishDelayThresholdMs: 200 # Default: 200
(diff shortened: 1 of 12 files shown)


core/src/main/java/com/spotify/ffwd/serializer/BuiltInSerializers.java
```diff
@@ -32,6 +32,7 @@
     public void setup() throws Exception {
         context.registerSerializer("spotify100", Spotify100Serializer.class);
         context.registerSerializer("spotify100proto", Spotify100ProtoSerializer.class);
+        context.registerSerializer("spotify100protobatch", Spotify100ProtoBatchSerializer.class);
```
hexedpackets: Why have this as a separate class instead of adding the method to `Spotify100ProtoSerializer`? You had to add a dummy implementation there anyway, and logically it makes sense to just have classes for the two serialization methods.

dmichel1: I was going back and forth on how to expose the batching to the user. I originally had everything as part of the `Spotify100ProtoSerializer` and added a configuration option to say batching and compression  should be enabled.

This [conditional](https://github.com/spotify/ffwd/pull/99/commits/77b429238bee0da10441462bc2cecef98e2c2ce2#diff-c9e098a35ccae1d31aa3f063c7ae61eaR115) would use that configuration option instead.

But then when I went to implement this on the receiving side in heroic it got confusing since the schema only [consumes bytes](https://github.com/spotify/heroic/blob/master/heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100.java#L154). So I chose to be specific how things should be serialized/deserialized rather than have another configuration option. 

I'm open to alternatives on implementation.

hexedpackets: For simplification, we could remove the single-metric from the protobuf implementation and only support batches. Sending a single metric would just be a batch of one then. Not sure if that's ideal, but it reduces the number of things to handle on both sides of the pipeline.

dmichel1: 👍 this is similar to what is already being done when batching is not enabled for the output plugin. 

```
  public void sendMetric(final Metric metric) {
    sendMetrics(Collections.singletonList(metric));
  }
```

I modified the existing proto schema to only support batches and changed the pubsub output plugin to only send batches. This will be a "breaking" change for the pubsub plugin.

dmichel1:
@hexedpackets updated if you could PTAL.

dmichel1:
I moved the 3 unrelated commits to a different pull request to get them released sooner - https://github.com/spotify/ffwd/pull/101 - will rebase this one after its merged

PR #105, opened 2019-01-15
Java11 update

sjoeboo:


dmichel1:
@sjoeboo i think we have to make ffwd work on java 11 first - someone tried to build under travis and it wasn't working as part of https://github.com/spotify/ffwd/pull/94 - I think we should merge something similar to this first.

lmuhlha:
This was done in https://github.com/spotify/ffwd/pull/107

PR #106, opened 2019-01-17
Add support to SignalFX output to allow batch input

jsferrei:


sjoeboo (approved):
looks good from here.

PR #110, opened 2019-03-07
Update java path and dependency to match internal requirements

lmuhlha:


pom.xml
```diff
@@ -109,7 +109,7 @@
               <dependency>
                 <groupId>com.puppycrawl.tools</groupId>
                 <artifactId>checkstyle</artifactId>
-                <version>8.11</version>
+                <version>[8.18,)</version>
```
dmichel1: is this version valid? I would keep the security update separate from the commit that updates the java path. They should be two atomic changes.

lmuhlha: Yeah when i tried it as just "8.18" it gave me issues, which was weird. This format ended up working fine. Ran tests and all that to confirm. Can separate the commits tho.

lmuhlha: Might've just been an intelli-j issue. Seemed to work alright now. Made a new commit.

dmichel1 (approved):

PR #112, opened 2019-03-19
Add revision version minimum for java install as part of the debian p…

lmuhlha:
…ackage

### DIFF
--- a/debian/control.in
+++ b/debian/control.in
@@ -10,6 +10,6 @@ Standards-Version: 3.9.1
 
 Package: ffwd-RELEASE
 Architecture: all
-Depends: oracle-jvm-11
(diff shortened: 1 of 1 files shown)


hexedpackets:
I don't think having an internal-specific build in the public debian package makes sense. I don't have dpkg locally to test, but would changing it to just `>=11.0.2` work as intended?

dmichel1:
i don't think it will since `>=11.0.2` is in the internal repo already. `oracle-jvm-11` is already a spotify-ism. 

I think this is a good reason to move this debian packaging into our internal deploy repo?

hexedpackets:
@dmichel1 agreed, if we can't make the package definition generic then it shouldn't be part of the OSS repo.

PR #118, opened 2019-04-04
Remove release notes

dmichel1:
Remove the release notes since we don't upload any of the artifacts to the maven repo.

This section should be added back with the correct release procedure when we have one.

@lmuhlha @hexedpackets


lmuhlha (approved):

hexedpackets:
There are a few places where we need to update the version manually. I think its also very helpful to have a git tag set for a release. Can you add these steps to the docs if we aren't going to use the plugin?

dmichel1:
Moving the new release process to https://github.com/spotify/ffwd/pull/137

PR #129, opened 2019-04-29
Add an expiring writeCache

dmichel1:
An ExpiringCache is useful to reduce the number of metrics being published to a topic. Typically processing and de-deduplication of the "metadata" around a metric only needs to happen every so often. In the case of using Heroic, this "metadata" only needs to be published at the interval of a new index being created.

@sjoeboo @lmuhlha @hexedpackets

### DIFF
--- a/api/src/main/java/com/spotify/ffwd/cache/ExpiringCache.java
+++ b/api/src/main/java/com/spotify/ffwd/cache/ExpiringCache.java
@@ -41,7 +41,7 @@ public class ExpiringCache implements WriteCache {
   }
 
   @Override
-  public boolean isCached(final Metric metric) {
+  public boolean checkCacheOrSet(final Metric metric) {
     if (writeCache.getIfPresent(metric.generateHash()) != null) {
       log.trace("Metric in cache: {}", metric);
       return true;
--- a/api/src/main/java/com/spotify/ffwd/cache/ExpiringCache.java
+++ b/api/src/main/java/com/spotify/ffwd/cache/ExpiringCache.java
@@ -1,0 +1,53 @@
+/*-
+ * -\-\-
+ * FastForward API
+ * --
+ * Copyright (C) 2016 - 2019 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ * -/-/-
+ */
+
+
+package com.spotify.ffwd.cache;
+
+import com.google.common.cache.Cache;
+import com.spotify.ffwd.model.Metric;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * ExpiringCache is useful to reduce the number of metrics being published to a topic. Typically
+ * processing and de-deduplication of the "metadata" around a metric only needs to happen every so
+ * often. In the case of using Heroic, this "metadata" only needs to be published at the interval of
+ * a new index being created.
+ */
+public class ExpiringCache implements WriteCache {
+  private static final Logger log = LoggerFactory.getLogger(ExpiringCache.class);
+  private final Cache<String, Boolean> writeCache;
+
+  public ExpiringCache(final Cache<String, Boolean> writeCache) {
+    this.writeCache = writeCache;
+  }
+
+  @Override
+  public boolean isCached(final Metric metric) {
+    if (writeCache.getIfPresent(metric.generateHash()) != null) {
+      log.trace("Metric in cache: {}", metric);
+      return true;
+    }
+    writeCache.put(metric.generateHash(), true);
+    return false;
+  }
+
+}
--- a/api/src/main/java/com/spotify/ffwd/model/Metric.java
+++ b/api/src/main/java/com/spotify/ffwd/model/Metric.java
@@ -20,6 +20,10 @@
 
 package com.spotify.ffwd.model;
 
+import com.google.common.base.Charsets;
+import com.google.common.hash.HashFunction;
+import com.google.common.hash.Hasher;
+import com.google.common.hash.Hashing;
 import java.util.Date;
 import java.util.Map;
 import java.util.Set;
@@ -32,6 +36,8 @@ import lombok.EqualsAndHashCode;
 @Data
 @EqualsAndHashCode(of = {"key", "riemannTags", "tags"})
 public class Metric {
+    static final HashFunction HASH_FUNCTION = Hashing.murmur3_128();
+
     private final String key;
     private final double value;
     private final Date time;
@@ -48,4 +54,28 @@ public class Metric {
     public Batch.Point toBatchPoint() {
         return new Batch.Point(key, tags, resource, value, time.getTime());
     }
+
+    public String generateHash() {
+        final Hasher hasher = HASH_FUNCTION.newHasher();
+
+        if (key != null) {
+            hasher.putString(key, Charsets.UTF_8);
+        }
+
+        for (final Map.Entry<String, String> kv : tags.entrySet()) {
+            final String k = kv.getKey();
+            final String v = kv.getValue();
+
+            if (k != null) {
+                hasher.putString(k, Charsets.UTF_8);
+            }
+
+            if (v != null) {
+                hasher.putString(v, Charsets.UTF_8);
+            }
+        }
+
+        return hasher.hash().toString();
+    }
+
 }
--- a/api/src/main/java/com/spotify/ffwd/model/Metric.java
+++ b/api/src/main/java/com/spotify/ffwd/model/Metric.java
@@ -43,8 +43,8 @@ public class Metric {
     private final double value;
     private final Date time;
     private final Set<String> riemannTags;
-    private final TreeMap<String, String> tags;
-    private final TreeMap<String, String> resource;
+    private final Map<String, String> tags;
+    private final Map<String, String> resource;
     private final String proc;
 
     /**
@@ -62,8 +62,9 @@ public class Metric {
         if (key != null) {
             hasher.putString(key, Charsets.UTF_8);
         }
+        TreeMap<String, String> sortedTags = new TreeMap<>(tags);
 
-        for (final Map.Entry<String, String> kv : tags.entrySet()) {
+        for (final Map.Entry<String, String> kv : sortedTags.entrySet()) {
             final String k = kv.getKey();
             final String v = kv.getValue();
 
--- a/api/src/main/java/com/spotify/ffwd/model/Metric.java
+++ b/api/src/main/java/com/spotify/ffwd/model/Metric.java
@@ -27,6 +27,7 @@ import com.google.common.hash.Hashing;
 import java.util.Date;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeMap;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
 
@@ -42,8 +43,8 @@ public class Metric {
     private final double value;
     private final Date time;
     private final Set<String> riemannTags;
-    private final Map<String, String> tags;
-    private final Map<String, String> resource;
+    private final TreeMap<String, String> tags;
+    private final TreeMap<String, String> resource;
     private final String proc;
 
     /**
--- a/docs/content/_docs/pubsub.md
+++ b/docs/content/_docs/pubsub.md
@@ -22,14 +22,33 @@ based on request size, message count and time since last publish. The client by 
 
 
 Here is an example config of setting up the pubsub plugin.
+
 ```
 output:
   plugins:
     - type: pubsub
-      flushInterval: 10000
       project: google-test-project
       topic: test-topic
+      serializer:
+        type: spotify100proto
+      
+      batching:
+        flushInterval: 10000
+        batchSizeLimit: 10000  # Default: 10000
+        maxPendingFlushes: 10 # Default: 10 
 ```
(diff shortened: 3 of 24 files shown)


hexedpackets:
@dmichel1 fyi you can use @spotify/layer8 to ping everyone instead of listing out people

api/src/main/java/com/spotify/ffwd/model/Metric.java
```diff
@@ -48,4 +54,28 @@
...
     }
+
+    public String generateHash() {
+        final Hasher hasher = HASH_FUNCTION.newHasher();
+
+        if (key != null) {
+            hasher.putString(key, Charsets.UTF_8);
+        }
+
+        for (final Map.Entry<String, String> kv : tags.entrySet()) {
```
hexedpackets: Is the order for entrySet guaranteed to be the same every time its called? I know it will for HashMaps, but I'm not sure about other Map implementations.

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
```diff
@@ -44,7 +45,7 @@
 import java.util.List;
 import java.util.UUID;
 import java.util.concurrent.Executor;
-import lombok.extern.slf4j.Slf4j;
+import org.slf4j.Logger;
```
hexedpackets, line 51: :yellow_heart:

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
```diff
@@ -82,52 +88,46 @@ public void init() { }
...
+        if (writeCache.isCached(metric)) {
+          publish = ApiFutures.immediateFuture("Metric in cache");
+        } else {
+          publish = publisher.publish(PubsubMessage.newBuilder()
+            .setData(ByteString.copyFrom(serializer.serialize(metric)))
+            .build()
+          );
+        }
+
         ApiFutures.addCallback(publish, new ApiFutureCallback<String>() {
```
hexedpackets: Since the async callbacks only are relevant to real publishes/cache misses, it would be easier to read if you add the publishPubSub function back and use that (with the cache changes of course). The loop in sendMetrics is starting to get large and messy.

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/cache/ExpiringCache.java
```diff
@@ -0,0 +1,56 @@
...
+  private final Cache<String, Boolean> writeCache;
+
+  public ExpiringCache(final Cache<String, Boolean> writeCache) {
+    this.writeCache = writeCache;
+  }
+
+  @Override
+  public boolean isCached(final Metric metric) {
+    if (writeCache.asMap().containsKey(metric.generateHash())) {
+      log.trace("Metric in cache: {}", metric);
```
hexedpackets: Having a metric generated by ffwd for the amount of cache hits/misses would be useful.

hexedpackets (commented):
Won't the cache drop all matching metrics until it expires, and not just the metadata?

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/cache/ExpiringCache.java
dmichel1: The guava cache provides statistics I can try to enable and convert to semantic metrics - https://github.com/google/guava/wiki/CachesExplained#statistics

api/src/main/java/com/spotify/ffwd/model/Metric.java
dmichel1: Good question. Looking at the `ProtobufDecoder` and the `JsonObjectMapperDecoder` the Metric is constructed with a `HashMap<>`. 

Is this a case where it would be better to be specific about the map implemented required or sort the keys before iterating over them?

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
dmichel1: Oh the `publishPubSub` function wasn't being used, that's why I deleted it. That might have been left over from when we reverted the protobuf change(?). Anyway I can refactor this method.

dmichel1:
@hexedpackets my thinking around how this feature would be used is to publish to two topics - e.g. `metrics-$region` and `metadata-$region`. Only the metadata topic would have this feature enabled.

api/src/main/java/com/spotify/ffwd/model/Metric.java
hexedpackets: I think so, might be good to change the type for `tags` to a HashMap if we can't confirm the general case for Map.

hexedpackets:
> @hexedpackets my thinking around how this feature would be used is to publish to two topics - e.g. `metrics-$region` and `metadata-$region`. Only the metadata topic would have this feature enabled.

Cool that makes sense. We should have that setup explained in the ffwd docs. If you don't want to add it as part of this PR, can you make a new issue for us to do later?

api/src/main/java/com/spotify/ffwd/model/Metric.java
dmichel1: I'm changing this to `TreeMap` to guarantee iteration order since we have little control over the various input plugins creating `Metric` objects

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
dmichel1: added this back.

dmichel1 (commented):
I added documentation. Need to add metrics and then will ping you.

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
```diff
@@ -81,18 +97,24 @@ public void init() { }
...
+    logger.debug("Sending events is not supported!");
     return async.resolved();
   }
 
-  private void publishPubSub(ByteString bytes) {
-    final ApiFuture<String> publish = publisher.publish(
-      PubsubMessage.newBuilder().setData(bytes).build());
+  private void publishPubSub(final ByteString bytes) {
+    // don't publish "\000" - indicates all the metrics are in the writeCache
+    if (bytes.size() <= 1) {
```
dmichel1, line 106: open to a better way to do this.

dmichel1:
@hexedpackets this is ready for another review pls.

api/src/main/java/com/spotify/ffwd/model/Metric.java
hexedpackets: Did you forget to commit the `TreeMap` change?

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubOutputPlugin.java
```diff
@@ -90,6 +107,11 @@ public PubsubOutputPlugin(
     this.project = ofNullable(project);
     this.topic = ofNullable(topic);
 
+    this.writeCacheEnabled = ofNullable(writeCacheEnabled).orElse(DEFAULT_CACHE_ENABLED);
```
hexedpackets: I'm generally not a fan of having to set an enabled flag in addition to configuring values. Can we just mark it enabled if either writeCacheDurationMinutes or writeCacheMaxSize is set (and use the default for the other one)?

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
```diff
@@ -69,6 +73,18 @@
   @Inject
   ProjectTopicName topicName;
 
+  @Inject
+  WriteCache writeCache;
+
+  @Inject
+  Logger logger;
```
hexedpackets, line 80: Injecting the logger seems a little weird, why not just create a static variable?

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
hexedpackets: Seems ok

modules/pubsub/src/test/com/spotify/ffwd/pubsub/PubsubPluginSinkTest.java
```diff
@@ -39,38 +46,56 @@ public void setUp() throws Exception {
...
-    private Metric makeMetric() {
-        return new Metric("key1", 1278, new Date(), ImmutableSet.of(),
-                ImmutableMap.of("what", "test"), ImmutableMap.of(), null);
-    }
 
     @Test
     public void testSendEvent() {
         sink.sendEvent(makeEvent());
-        verify(publisher).publish(any(PubsubMessage.class));
+        verifyZeroInteractions(publisher);
```
hexedpackets, line 62: What did you change that would make this have zero interactions?

docs/content/_docs/pubsub.md
```diff
@@ -22,16 +22,35 @@ based on request size, message count and time since last publish. The client by
...
+        flushInterval: 10000
+        batchSizeLimit: 10000  # Default: 10000
+        maxPendingFlushes: 10 # Default: 10 
 ```
 
 
+### Write Cache
+
+There is an optional write cache that can be enabled. At Spotify this is being used as a "cheap" method
+to deduplicate the metadata around timeseries when using the heroic tsbd. The way this feature is used is by publishing from ffwd twice.
```
hexedpackets: title case - `Heroic TSDB`

api/src/main/java/com/spotify/ffwd/cache/ExpiringCache.java
```diff
@@ -0,0 +1,53 @@
...
+public class ExpiringCache implements WriteCache {
+  private static final Logger log = LoggerFactory.getLogger(ExpiringCache.class);
+  private final Cache<String, Boolean> writeCache;
+
+  public ExpiringCache(final Cache<String, Boolean> writeCache) {
+    this.writeCache = writeCache;
+  }
+
+  @Override
+  public boolean isCached(final Metric metric) {
```
hexedpackets: I suggest renaming the method - in my mind something named `isCached` would be a side-effect free check.

hexedpackets:
A couple small suggestions and questions but looking good!

api/src/main/java/com/spotify/ffwd/model/Metric.java
dmichel1: something got messed up. adding the `TreeMap` back.

api/src/main/java/com/spotify/ffwd/cache/ExpiringCache.java
dmichel1: `isCachedOrSet`? Blanking on a better name.

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubOutputPlugin.java
dmichel1: good suggestion - we do this in other places already. will enable if `writeCacheDurationMinutes` is non-zero

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
dmichel1: I found out there is https://github.com/spotify/ffwd/blob/master/api/src/main/java/com/spotify/ffwd/output/OutputPluginModule.java#L39 which can be used.

Now that there will be two pubsub output plugins having the log message have the pluginId is useful.

I don't care if we get rid of the injection and create the logger directly. Would need to inject `pluginId`.

modules/pubsub/src/test/com/spotify/ffwd/pubsub/PubsubPluginSinkTest.java
dmichel1: We don't publish events since its not supported.

api/src/main/java/com/spotify/ffwd/cache/ExpiringCache.java
hexedpackets: `checkCacheOrSet`? `maybeCache`?

modules/pubsub/src/test/com/spotify/ffwd/pubsub/PubsubPluginSinkTest.java
hexedpackets: Still confused why that would have published before and now it doesn't.

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
hexedpackets: Cool lets leave that then. But what's the second pubsub plugin? We still only have one afaict.

hexedpackets (approved):
The `isCached` is the only thing remaining (and I had two questions for my own knowledge that don't need any code change) so I'm LGTMing.

api/src/main/java/com/spotify/ffwd/cache/ExpiringCache.java
dmichel1: changing it to checkCacheOrSet

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
dmichel1: it's the same pubsub module, but we create two output plugins for the different topics.

modules/pubsub/src/test/com/spotify/ffwd/pubsub/PubsubPluginSinkTest.java
dmichel1: ah I don't remember, what the problem was, wrote this too long ago. I guess test itself was broken/not actually testing the right thing.

Issue #131, opened 2019-05-02
signalfx output plugin reports all metrics as gauges


dmichel1:
Currently the signalfx output plugin reports all metrics as gauges instead of the appropriate metric type

https://github.com/spotify/ffwd/blob/c6628498deaa66553205fe7399963b3d5a2c9423/modules/signalfx/src/main/java/com/spotify/ffwd/signalfx/SignalFxPluginSink.java#L111

This causes default rollups and aggregations in Siganlfx to be incorrect for cumulative counters. And also produced negative values when instances are reset.

https://docs.signalfx.com/en/latest/charts/resolution-rollups.html#rollup-types
>Rate/sec (default for counter and cumulative counter metrics): Divide the result of Sum (counter) or Delta (cumulative counter) by the number of seconds in the time interval.

PR #134, opened 2019-05-06
[signalfx] output counters as correct metric type

dmichel1:
Currently all metrics are being sent to SignalFx as a gauge. This makes default rollups and aggregations
not handle counters correctly as when they are reset they will produce negative values.

https://docs.signalfx.com/en/latest/getting-started/concepts/metric-types.html
https://docs.signalfx.com/en/latest/charts/resolution-rollups.html#rollups-in-more-detail


dmichel1:
Fixes https://github.com/spotify/ffwd/issues/131

hexedpackets (approved):

PR #138, opened 2019-05-09
Include version in debian package name

dmichel1:
this matches how the previous debian package was being named.

### DIFF
--- a/agent/pom.xml
+++ b/agent/pom.xml
@@ -136,7 +136,7 @@
         <configuration>
           <controlDir>${basedir}/src/deb/control</controlDir>
           <verbose>true</verbose>
-          <deb>${project.build.directory}/ffwd.deb</deb>
+          <deb>${project.build.directory}/ffwd_${version}_all.deb</deb>
           <snapshotExpand>true</snapshotExpand>
           <!-- or defaults to a timestamp -->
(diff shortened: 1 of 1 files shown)


hexedpackets:
@dmichel1 We removed the version because it makes it difficult to upgrade or publish outside of Spotify. Having the package named just `ffwd` with an actual version instead of always being latest makes a lot more sense IMO.

dmichel1:
@hexedpackets I'm missing some debian knowledge here but in the control file the package is still `Package: ffwd` 

This is how the files look like in our internal debrepo...
```
ls -ltr debian-repository/trusty/pool/stable/main/ffwd/

ffwd_0.4.6_all.deb
ffwd_0.4.7_all.deb
```

vs the old package naming you are referencing...
```
ls -ltr debian-repository/trusty/pool/stable/main/ffwd-0.4.1
ffwd-0.4.1_1_all.deb
```

Right now with the actual package name being `ffwd` I can never see it via `apt-cache policy ffwd`

hexedpackets:
@dmichel1 ah you're right, I saw the change and didn't realize it was for the pom. Sorry!

hexedpackets (approved):

dmichel1:
all good - tack!

PR #140, opened 2019-05-13
[http] migrate from lombok to kotlin

dmichel1:
Mostly doing this to a fix a bug with java11 and lombok when deserialization the yaml config into the HttpDiscovery object.

```
Caused by: com.fasterxml.jackson.databind.exc.MismatchedInputException: Cannot construct instance of `com.spotify.ffwd.http.HttpDiscovery$Srv` (although at least one Creator exists): cannot deserialize from Object value (no delegate- or property-based Creator)
at [Source: (sun.nio.ch.ChannelInputStream); line: 66, column: 7] (through reference chain: com.spotify.ffwd.AgentConfig["output"]->com.spotify.ffwd.output.OutputManagerModule["plugins"]->java.util.ArrayList[0]->com.spotify.ffwd.http.HttpOutputPlugin["discovery"])
at com.fasterxml.jackson.databind.exc.MismatchedInputException.from(MismatchedInputException.java:63)
```


agent/pom.xml
```diff
@@ -128,15 +123,14 @@
...
-
       <plugin>
         <artifactId>jdeb</artifactId>
         <groupId>org.vafer</groupId>
         <version>1.7</version>
         <configuration>
           <controlDir>${basedir}/src/deb/control</controlDir>
           <verbose>true</verbose>
-          <deb>${project.build.directory}/ffwd_${version}_all.deb</deb>
+          <deb>${project.build.directory}/ffwd_${project.version}_all.deb</deb>
```
dmichel1, line 133: using just `${version}` is deprecated.

core/src/main/java/com/spotify/ffwd/domain/SearchDomainDiscovery.kt
```diff
@@ -0,0 +1,120 @@
...
+            }
+
+        }
+    }
+
+    companion object {
+        /**
+         * Default implementation for http discovery when nothing else is configured.
+         */
+        fun supplyDefault(): SearchDomainDiscovery {
```
hexedpackets, line 117: You can add the annotation `@JvmStatic` and access this the same as the Java static function, `SearchDomainDiscovery::supplyDefault`, instead of having to reference `.Companion`.

modules/http/src/main/java/com/spotify/ffwd/http/HttpClient.java
```diff
@@ -36,16 +35,23 @@
...
-    public static final String PING_ENDPOINT = "ping";
+    private static final String V1_BATCH_ENDPOINT = "v1/batch";
+    private static final String PING_ENDPOINT = "ping";
 
     private final AsyncFramework async;
     private final ObjectMapper mapper;
     private final OkHttpClient httpClient;
     private final String baseUrl;
 
+    public HttpClient(AsyncFramework async, ObjectMapper mapper, OkHttpClient httpClient,
```
hexedpackets, line 47: I think you'll need to add jackson annotations on the constructor. That's something the lombok `Data` classes do invisibly.

modules/http/src/main/java/com/spotify/ffwd/http/HttpDiscovery.kt
```diff
@@ -0,0 +1,101 @@
...
+
+
+    @JsonTypeName("static")
+    data class Static(val servers: List<HostAndPort>) : HttpDiscovery {
+
+        override fun apply(
+                builder: LoadBalancerBuilder<Server>, searchDomain: Optional<String>
+        ): LoadBalancerBuilder<Server> {
+            val servers = this.servers
+                    .stream()
```
hexedpackets: I think you can remove the stream and collect here, kotlin lets you do functional mappings on the list directly.

modules/http/src/main/java/com/spotify/ffwd/http/HttpException.kt
```diff
@@ -18,17 +18,8 @@
...
-@Data
-public class HttpException extends RuntimeException {
-    private final HttpResponseStatus status;
-
-    public HttpException(final HttpResponseStatus status) {
-        super(status.reasonPhrase());
-        this.status = status;
-    }
-}
+class HttpException(val status: HttpResponseStatus?) : RuntimeException(status?.reasonPhrase())
```
hexedpackets: The Java version looks like it was non-nullable (or at least it would throw a null-pointer exception on the super call), so I suggest making the Kotlin param non-nullable to match it.

modules/http/src/main/java/com/spotify/ffwd/http/SrvServerList.kt
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        val result = lookup.run()
+
+        if (lookup.result != Lookup.SUCCESSFUL) {
+            throw RuntimeException(
+                    "DNS lookup failed: " + lookup.errorString + ": " + record)
+        }
+
+        val results = ImmutableList.builder<Server>()
```
hexedpackets, line 54: Kotlin lists are immutable by default, so the guava list could be removed from this class.

modules/http/src/main/java/com/spotify/ffwd/http/HttpException.kt
dmichel1: added a `!!` null check 👍

core/src/main/java/com/spotify/ffwd/domain/SearchDomainDiscovery.kt
dmichel1: oh right thanks for the reminder, I remember you doing that in heroic... fixed

modules/http/src/main/java/com/spotify/ffwd/http/HttpClient.java
dmichel1: This class is only created through the HttpClientFactory. I don't think the lombok data class was ever needed.

modules/http/src/main/java/com/spotify/ffwd/http/HttpDiscovery.kt
dmichel1: TIL - and made this just `toList()`

modules/http/src/main/java/com/spotify/ffwd/http/SrvServerList.kt
dmichel1: hmm but I have to build a list here...

```
        val results : MutableList<Server> = ArrayList()

        if (result != null) {
            for (a in result) {
                val srv = a as SRVRecord
                results.add(Server(a.target.canonicalize().toString(), srv.port))
            }
        }

        return results
```

will return a mutable list right?

modules/http/src/main/java/com/spotify/ffwd/http/HttpClient.java
hexedpackets: :+1:

modules/http/src/main/java/com/spotify/ffwd/http/HttpException.kt
hexedpackets: So what your class says now is that it takes a nullable HttpResponseStatus as the constructor arg, then supers a RuntimeException with a non-null check. You can remove the !! and change `HttpResponseStatus?` to `HttpResponseStatus` and the null check will happen at compile time.

dmichel1: fixed.

modules/http/src/main/java/com/spotify/ffwd/http/SrvServerList.kt
hexedpackets: Ah right. So yeah I think sticking with the guava list is fine, or for the just-Kotlin approach you could do the MutableList and then cast it to List before the return to make it immutable - `val immutableResults: List<Server> = results`. Either way works, and both avoid returning a mutable list.

dmichel1: I think i prefer the builder pattern - going to leave guava there - found out about https://github.com/Kotlin/kotlinx.collections.immutable too

hexedpackets (approved):

Issue #132, opened 2019-05-03
Enforce rate limiting


hexedpackets:
Having a per-host rate limit will protect the rest of the metric pipeline against someone adding a bad metric.

There should be a metric when things are being rate-limited that way it can be monitored for.

Can do this guava - https://google.github.io/guava/releases/19.0/api/docs/index.html?com/google/common/util/concurrent/RateLimiter.html.

The limiting can be added where the current filter logic is done or as another filter itself (i think the latter is better).

dmichel1:
Thinking about how this is deployed internally... 

With GCE we can utilize puppet to allow teams to override the rate-limit.

With k8s we need to investigate if setting an environment variable will work with how the ffwd side car is injected with the admission controller. If it doesn't we might have to provide another mechanism to do so.

hexedpackets:
With admission controllers, the whole container object is injected by the controller and can't be changed by the deployment.

We could use annotations on the pod, and have ffwd read those on startup if its running in k8s. That would also be extendable to setting limits on a namespace instead of every single pod.

hexedpackets:
Instead of reading the k8s annotation in ffwd itself, we could use the downward API. It would require a bit more setup for someone using ffwd in kubernetes but not much and removes any issues around permissions for the kubernetes API. 

For example, with an annotation of `ffwd.spotify.com/rate: 1k` the container can then be configured with
```yaml
env:
- name: FFWD_RATE_LIMIT
  valueFrom:
    fieldRef:
      fieldPath: metadata.annotations['ffwd.spotify.com/rate']
```
This seems like a pretty good solution IMO. Any thoughts @dmichel1 ?

dmichel1:
+1 it's really simple to customize for users.

PR #147, opened 2019-07-09
Update configuration to load from yaml or env vars.

### LINKED ISSUE #120: Ignore unknown config fields
Right now, unknown config fields will cause a runtime exception. We should ignore these to make it easier to update the config when rolling out new versions.

hexedpackets:
Env vars are the preferred way to configure ffwd when its running in kubernetes, and will be needed for #132 

This also fixes #120 since the Konf library ignores unknowns by default.

### DIFF
--- a/agent/pom.xml
+++ b/agent/pom.xml
@@ -92,6 +92,11 @@
       <artifactId>mockito-core</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>com.github.stefanbirkner</groupId>
+      <artifactId>system-rules</artifactId>
+      <scope>test</scope>
+    </dependency>
 
     <!-- semantic metrics -->
     <dependency>
@@ -119,6 +124,18 @@
             <goals>
               <goal>shade</goal>
             </goals>
+            <configuration>
+              <filters>
+                <filter>
+                  <artifact>*:*</artifact>
+                  <excludes>
+                    <exclude>META-INF/*.SF</exclude>
+                    <exclude>META-INF/*.DSA</exclude>
+                    <exclude>META-INF/*.RSA</exclude>
+                  </excludes>
+                </filter>
+              </filters>
+            </configuration>
           </execution>
         </executions>
       </plugin>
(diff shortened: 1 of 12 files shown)


hexedpackets:
We don't currently have docs for the configuration. I made a new issue to address that: #148

core/src/main/java/com/spotify/ffwd/AgentConfig.kt
```diff
@@ -16,53 +16,83 @@
...
+    val debugLocalAddress =  config[Debug.localAddress]
+    val host = config[AgentConfig.host]
+    val tags = config[AgentConfig.tags]
+    val tagsToResource = config[AgentConfig.tagsToResource]
+    val riemannTags = config[AgentConfig.riemannTags]
+    val skipTagsForKeys = config[AgentConfig.skipTagsForKeys]
+    val automaticHostTag = config[AgentConfig.automaticHostTag]
+    val input: InputManagerModule = config[AgentConfig.input]
+    val output: OutputManagerModule = config[AgentConfig.output]
+    val searchDomain = config[AgentConfig.searchDomain]
```
lmuhlha, line 45: should this be SearchDomainDiscovery?

agent/pom.xml
```diff
@@ -119,6 +124,18 @@
             <goals>
               <goal>shade</goal>
             </goals>
+            <configuration>
```
dmichel1, line 127: what is this for?

core/src/main/java/com/spotify/ffwd/AgentConfig.kt
hexedpackets: The type is `SearchDomainDiscovery` but I kept the original variable name of `searchDomain`.

dmichel1: the object type still is - down below the config will return ` val searchDomain by lazy { SearchDomainDiscovery.supplyDefault() }`

dmichel1 (approved):
1 general comment about the shading - lgtm

core/src/main/java/com/spotify/ffwd/AgentConfig.kt
lmuhlha: 👍

agent/pom.xml
hexedpackets: There are apparently signed jar dependencies that cause the shading to fail with a security error. The filter removes the signed artifacts from the jars its bundling so that the JVM will actually run it.

Issue #148, opened 2019-07-09
Document configuration


hexedpackets:
The documentation should have all of the configuration fields, and the ways to load them (yaml file, env vars, etc).

PR #149, opened 2019-07-10
Enable rate limiting in the output manager.

### LINKED ISSUE #132: Enforce rate limiting

hexedpackets:
By default rate limiting is disabled. It can be configured in the yaml config as 
```yaml
output:
  rateLimit: x
```
or with the environment variable `FFWD_OUTPUT_RATELIMIT`. The value is number of metrics per second to allow. The limiter is non-blocking and metrics that exceed the limit will be dropped.

While exploring the output code I simplified some of the batching deserialization logic. Its a fairly small change, but I'm happy to break that out into a separate PR since its not really related.

Closes #132

### DIFF
--- a/api/src/main/java/com/spotify/ffwd/module/Batching.java
+++ b/api/src/main/java/com/spotify/ffwd/module/Batching.java
@@ -23,13 +23,14 @@ package com.spotify.ffwd.module;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import java.util.Optional;
+import javax.annotation.Nullable;
 import lombok.Data;
 
 @Data
 public class Batching {
     public static final boolean DEFAULT_REPORT_STATISTICS = false;
 
-    protected final Optional<Long> flushInterval;
+    @Nullable protected final Long flushInterval;
     protected final Optional<Long> batchSizeLimit;
     protected final Optional<Long> maxPendingFlushes;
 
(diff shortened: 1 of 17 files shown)


api/src/main/java/com/spotify/ffwd/module/Batching.java
```diff
@@ -74,34 +65,15 @@ public static Batching from(
      * This only contains something when the user didn't use any 'batching' sub structure in the
      * configuration, and just specified flushInterval.
      * @param batching A complete Batching structure, on the output plugin level in the conf.
-     * @param defaultFlushInterval Optional default value to be used if no flushInterval nor
      * batching was specified.
```
lmuhlha: remove this line too

lmuhlha (approved):

dmichel1:
Can we add something to the docs about this? Also one thing worth mentioning there perhaps is the rate limiter is shared between events and metrics. I don't think it matters much in practice, especially with how we use FFWD internally.

hexedpackets:
ack

hexedpackets:
The ffwd docs are pretty weak right now. I was planning to pick up #148 soon and document the rate limiting there (along with all the other config options).

PR #150, opened 2019-07-10
Upgrade jackson.

hexedpackets:
This patch resolves two open CVEs against older jackson-databind versions.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -236,35 +236,17 @@
         <version>0.13.3</version>
       </dependency>
 
-      <!-- Jackson libraries -->
       <dependency>
-        <groupId>com.fasterxml.jackson.core</groupId>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -236,13 +236,17 @@
         <version>0.13.3</version>
       </dependency>
 
+      <!-- Jackson libraries -->
+      <dependency>
+        <groupId>com.fasterxml.jackson.core</groupId>
+        <artifactId>jackson-core</artifactId>
+        <version>${jackson.version}</version>
+      </dependency>
       <dependency>
```
dmichel1: we can import the jackson bom to cut down on all these dependencies. 

https://github.com/FasterXML/jackson-bom#usage

dmichel1 (approved):

Issue #130, opened 2019-04-29
Deprecate events api


dmichel1:
Currently the event support of ffwd is not used internally within Spotify and there are no plans to. Does it make sense to continue supporting it before we cut a major release of ffwd?

hexedpackets:
Its also experimental/unfinished on the Heroic side of things.

hexedpackets:
Since we've decided to remove events from Heroic (https://github.com/spotify/heroic/issues/479) we should also remove them from ffwd.

hexedpackets:
To fully remove this we'll have to also remove the riemann input, since that exclusively outputs events. 

@dmichel1 do you think we can do that now or do we need a slower deprecation? I don't think internally we're using that input anywhere. So maybe after #120 is done we can remove it.

dmichel1:
@hexedpackets are we 100%  on removing event support as a concept from ffwd? this decision can be made independently from heroic, as I don't think heroic is a good fit for this type of data. 

I could see a future where we want to have a more robust pipeline of events that go to elasticsearch or a similar database. Whether or not we want to funnel that through ffwd is the bigger question.

hexedpackets:
Events have never been fully supported - kafka and riemann were the only outputs that allowed them, everything else just drops events. I also think events are different enough from metrics that it doesn't make sense to funnel them through the same pipeline. Events are closer to logs with a need for stronger delivery guarantees. Guaranteed delivery isn't a goal of ffwd, and many of the interesting things we can do around metrics wouldn't apply to events or would have to be implemented very differently.

tl;dr: yes let's remove it

dmichel1:
@hexedpackets sgtm - lets remove it!

PR #153, opened 2019-07-11
Remove events to only support metrics.

hexedpackets:
Removes events completely from the pipeline with the exception of the riemann input.

#130

### DIFF
--- a/core/src/main/java/com/spotify/ffwd/debug/NettyDebugServer.java
+++ b/core/src/main/java/com/spotify/ffwd/debug/NettyDebugServer.java
@@ -123,7 +123,7 @@ public class NettyDebugServer implements DebugServer {
                 connected.add(ch);
                 log.info("Connected {}", ch);
 
-                ch.closeFuture().addListener((ChannelFutureListener) future1 -> {
+                ch.closeFuture().addListener((ChannelFutureListener) f -> {
                     connected.remove(ch);
                     log.info("Disconnected {}", ch);
                 });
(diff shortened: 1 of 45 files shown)


core/src/main/java/com/spotify/ffwd/debug/NettyDebugServer.java
```diff
@@ -132,38 +119,32 @@ private void sendInspectPacket(Object event) throws Exception {
...
                 connected.add(ch);
                 log.info("Connected {}", ch);
 
-                ch.closeFuture().addListener(new ChannelFutureListener() {
-                    @Override
-                    public void operationComplete(ChannelFuture future) throws Exception {
-                        connected.remove(ch);
-                        log.info("Disconnected {}", ch);
-                    }
+                ch.closeFuture().addListener((ChannelFutureListener) future1 -> {
```
dmichel1: call this `f` or `future`? you use `f` below.

modules/protobuf/src/main/java/com/spotify/ffwd/protobuf/ProtobufDecoder.java
```diff
@@ -145,18 +141,4 @@ private Object decodeMetric0(final Protocol0.Metric metric) {
 
         return attributes;
     }
-
-    private Object decodeEvent0(final Protocol0.Event event) {
```
dmichel1, line 149: ah we have to remove event from https://github.com/spotify/ffwd-client-java too

dmichel1 (approved):
1 nit. LGTM after discussing more in https://github.com/spotify/ffwd/issues/130

core/src/main/java/com/spotify/ffwd/debug/NettyDebugServer.java
hexedpackets: Renamed it to `f` (this was intellij's auto-refactoring btw).

modules/protobuf/src/main/java/com/spotify/ffwd/protobuf/ProtobufDecoder.java
hexedpackets: Created https://github.com/spotify/ffwd-client-java/issues/3 to do that

PR #158, opened 2019-08-02
Replace Guava RateLimiter with a bursty token bucket.

hexedpackets:
The Guava RateLimiter uses a smoothing function to evenly disperse its tokens. This is problematic for bursty metric ingestion, as is often the case if the service batches metrics before sending them.

The token bucket is much simpler. It allows tokens to be acquired without regard to timing, and then refills the tokens on a fixed interval tick. If a burst of traffic needs to use all available permits, the bucket will allow it and on the next tick (every 500ms as configured in this PR) more tokens will be added to the bucket.


malish8632 (approved):
LGTM

PR #159, opened 2019-08-02
Update maven license plugin.

hexedpackets:
Newer versions of the plugin allow for trimming whitespace on the header, and will allow both trimmed and untrimmed headers when checking if a file has the license applied.


hexedpackets:
The build is failing because the Travis maven proxy is broken.

malish8632 (approved):

PR #163, opened 2019-08-19
[generated] Fix host entry in resource map in GeneratedPluginSource.

hnjslater:
As far as I can tell, GeneratedPluginSource is currently broken. I'm not sure if I'm missing something obvious but here's a fix that got it working for me.


hexedpackets (approved):
Thanks for the fix! Looks like this broke in b37c0e594679e7daa40ed93baf6e3951b62c2e47 when host become a tag instead of a top-level attribute.

PR #175, opened 2019-11-13
Update build link

dmichel1:


hexedpackets (approved):

dmichel1:
@hexedpackets i can't any indication of the circle-build being trigged hmmm

dmichel1:
maybe because it doesn't match any of the branch names?

https://github.com/spotify/ffwd/blob/master/.circleci/config.yml#L10

PR #179, opened 2019-11-15
Add docker image

dmichel1:
The docker image will allow for quicker local troubleshooting and for easier deployments.

Dockerhub has been setup to build master as the latest image tag and any git tags as image tags.

I replaced the default ffwd.yaml with the local debugging configuration since it's more useful.


README.md
```diff
@@ -11,12 +11,40 @@ ffwd decorates the received metrics with system-wide tags or attributes. By doin
...
 __Head over to https://spotify.github.io/ffwd/ for documentation.__
 
-# Building
+
+# Quick Start
+
+ffwd can be started quickly with docker. This can be useful to run locally when troubleshooting metrics with your service.
+
+```bash
+docker run -it -p 19000:19000 -p 8080:8080 spotify/ffwd:latest
```
hexedpackets: Should we add the udp ports to the run command?

hexedpackets (approved):

README.md
dmichel1: good catch - i'll add `19091` since that is what apollo is using

PR #186, opened 2020-01-28
Cardinality detection and limiting implementation

### LINKED ISSUE #181: High cardinality detection to FFWD
High cardinality metrics are hard to manage and create unnecessary load on backend systems.
Earlier detection and following filtering of such metrics would help to automatically manage system without need for user intervention.

malish8632:
- Add cardinality limit based on [HyperLogLogPlus](https://github.com/addthis/stream-lib/blob/master/src/main/java/com/clearspring/analytics/stream/cardinality/HyperLogLogPlus.java)
- Add cardinality metric for observability
- For **batch** using "commonTags" and "commonResource" for hashing - feel this should be reviewed as we go to include all tags and values in the batch
- For single **metric** using same tags as original metric ("key", "riemannTags", "tags")
- Cardinality will be reset by swapping HyperLogLogPlus after configurable time in case it was tripped

Closes: #181

### DIFF
--- a/agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
+++ b/agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
@@ -90,7 +90,7 @@ public class FfwdConfigurationTest {
         environmentVariables.set("FFWD_TTL", "100");
         environmentVariables.set("FFWD_OUTPUT_RATELIMIT", "1000");
         environmentVariables.set("FFWD_OUTPUT_CARDINALITYLIMIT", "10000");
-        environmentVariables.set("FFWD_OUTPUT_HYPERLOGLOGSWAPPERIODMS", "3000000");
+        environmentVariables.set("FFWD_OUTPUT_CARDINALITYTTL", "3000000");
 
         CoreOutputManager outputManager = getOutputManager(null);
 
--- a/agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
+++ b/agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
@@ -90,7 +90,7 @@ public class FfwdConfigurationTest {
         environmentVariables.set("FFWD_TTL", "100");
         environmentVariables.set("FFWD_OUTPUT_RATELIMIT", "1000");
         environmentVariables.set("FFWD_OUTPUT_CARDINALITYLIMIT", "10000");
-        environmentVariables.set("FFWD_OUTPUT_HYPERLOGLOGSWAPPERIOD", "3000000");
+        environmentVariables.set("FFWD_OUTPUT_HYPERLOGLOGSWAPPERIODMS", "3000000");
 
         CoreOutputManager outputManager = getOutputManager(null);
 
--- a/agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
+++ b/agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
@@ -89,11 +89,13 @@ public class FfwdConfigurationTest {
     public void testConfigFromEnvVars() {
         environmentVariables.set("FFWD_TTL", "100");
         environmentVariables.set("FFWD_OUTPUT_RATELIMIT", "1000");
+        environmentVariables.set("FFWD_OUTPUT_CARDINALITYLIMIT", "10000");
 
         CoreOutputManager outputManager = getOutputManager(null);
 
         assertEquals(100, outputManager.getTtl());
         assertEquals(Long.valueOf(1000), outputManager.getRateLimit());
+        assertEquals(Long.valueOf(10000), outputManager.getCardinalityLimit());
     }
 
     @Test
@@ -112,6 +114,7 @@ public class FfwdConfigurationTest {
 
         assertEquals(100, outputManager.getTtl());
         assertEquals("jimjam", outputManager.getHost());
+        assertEquals(Long.valueOf(10000), outputManager.getCardinalityLimit());
     }
 
     private CoreOutputManager getOutputManager(final Path configPath) {
--- a/agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
+++ b/agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
@@ -90,12 +90,14 @@ public class FfwdConfigurationTest {
         environmentVariables.set("FFWD_TTL", "100");
         environmentVariables.set("FFWD_OUTPUT_RATELIMIT", "1000");
         environmentVariables.set("FFWD_OUTPUT_CARDINALITYLIMIT", "10000");
+        environmentVariables.set("FFWD_OUTPUT_HYPERLOGLOGSWAPPERIOD", "3000000");
 
         CoreOutputManager outputManager = getOutputManager(null);
 
         assertEquals(100, outputManager.getTtl());
         assertEquals(Long.valueOf(1000), outputManager.getRateLimit());
         assertEquals(Long.valueOf(10000), outputManager.getCardinalityLimit());
+        assertEquals(Long.valueOf(3000000), outputManager.getHLLPRefreshPeriodLimit());
     }
 
     @Test
@@ -115,6 +117,7 @@ public class FfwdConfigurationTest {
         assertEquals(100, outputManager.getTtl());
         assertEquals("jimjam", outputManager.getHost());
         assertEquals(Long.valueOf(10000), outputManager.getCardinalityLimit());
+        assertEquals(Long.valueOf(555555), outputManager.getHLLPRefreshPeriodLimit());
     }
 
     private CoreOutputManager getOutputManager(final Path configPath) {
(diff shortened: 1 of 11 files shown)


lmuhlha:
seems reasonable to me 👍

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -220,6 +292,22 @@ private boolean rateLimitAllowed(int permits, String key) {
...
     }
 
+    private boolean cardinalityLimitAllowed(long currentCardinality) {
+        return cardinalityLimit == null || cardinalityLimit >= currentCardinality;
+    }
+
+    /**
+    * To reset cardinality this will swap HLL++ if it was tripped after configured period of ms
+    */
+    private void swapHyperLogLogPlus() {
```
malish8632, line 273: @hexedpackets It looks ugly and I'm curious if there is other way to atomically swap the objects

core/src/main/java/com/spotify/ffwd/output/OutputManagerModule.java
```diff
@@ -56,16 +56,22 @@
...
+    @Nullable private final Long hyperLogLogPlusSwapPeriod;
 
     @JsonCreator
     public OutputManagerModule(
         @JsonProperty("plugins") List<OutputPlugin> plugins,
         @JsonProperty("filter") Filter filter,
-        @JsonProperty("ratelimit") @Nullable Integer rateLimit
+        @JsonProperty("ratelimit") @Nullable Integer rateLimit,
+        @JsonProperty("cardinalitylimit") @Nullable Long cardinalityLimit,
+        @JsonProperty("hyperloglogswapperiod") @Nullable Long hyperLogLogPlusSwapPeriod
```
dmichel1: nit; include time unit in name.

core/pom.xml
```diff
@@ -1,5 +1,17 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-compiler-plugin</artifactId>
```
dmichel1: did you mean to add this? why is this needed now. should be 11

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -133,6 +161,22 @@ public final Long getRateLimit() {
         } else {
             rateLimiter = null;
         }
+
+        hyperLog = new HyperLogLogPlus(14, 25);
```
dmichel1: what are these magic numbers? would be good to define them as constants.

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -176,10 +234,25 @@ public void sendBatch(Batch batch) {
...
         int batchSize = batch.getPoints().size();
 
-        if (batchSize > 0 && !rateLimitAllowed(batchSize, batch.getPoints().get(0).getKey())) {
-            log.debug("Dropping {} metrics due to rate limiting", batchSize);
-            statistics.reportMetricsDroppedByRateLimit(batchSize);
-            return;
+        hyperLog.offer(batch.hashCode());
+        statistics.reportMetricsCardinality(batch.hashCode());
+
+        if (!Arrays.asList(KEYS_NEVER_TO_DROP).contains(batch.getPoints().get(0).getKey())) {
```
dmichel1: this code for `batch` is nearly identical to the one for a single metric. can the code be deduplicated?

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -220,6 +292,22 @@ private boolean rateLimitAllowed(int permits, String key) {
...
+        return cardinalityLimit == null || cardinalityLimit >= currentCardinality;
+    }
+
+    /**
+    * To reset cardinality this will swap HLL++ if it was tripped after configured period of ms
+    */
+    private void swapHyperLogLogPlus() {
+        if (System.currentTimeMillis() - hyperLogSwapTS.get() > hyperLogLogPlusSwapPeriod
+            && hyperLogSwapLock.compareAndExchange(false, true)) {
+          hyperLog = new HyperLogLogPlus(14, 25);
```
dmichel1: i might be wrong but I think you need a `synchronized` bit here so all threads will see the new reference. i need to come back to this part.

core/src/main/resources/log4j2.xml
```diff
@@ -9,7 +9,7 @@
     <Logger name="kafka.consumer" level="INFO" additivity="false">
       <AppenderRef ref="primary" />
     </Logger>
-    <Logger name="com.spotify.ffwd" level="WARN" additivity="false">
+    <Logger name="com.spotify.ffwd" level="INFO" additivity="false">
```
dmichel1: does this produce a lot of logs? iirc we changed this to reduce the logging in production. perhaps you can add a logger just for the class you are interested in instead?

core/pom.xml
malish8632: nope, going to remove. Intellij gives me a headache with those modules and correct versions

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
malish8632: done

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -113,18 +118,41 @@ public long getTtl() {
...
             return null;
         }
         return rateLimiter.getCapacity();
     }
 
+    public final Long getCardinalityLimit() {
+        if (cardinalityLimit == null) {
+            return null;
+        }
+        return cardinalityLimit;
```
hexedpackets: This is redundant, you can just return `cardinalityLimit` without the null check.

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -113,18 +118,41 @@ public long getTtl() {
...
+    }
+
+    public final Long getHLLPRefreshPeriodLimit() {
+        if (hyperLogLogPlusSwapPeriod == null) {
+            return null;
+        }
+        return hyperLogLogPlusSwapPeriod;
```
hexedpackets: Same here

core/src/main/java/com/spotify/ffwd/output/OutputManagerModule.java
malish8632: done

core/src/main/resources/log4j2.xml
malish8632: reverting it back, just missed it

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
hexedpackets: hmm, I can't think of a cleaner way to do it

malish8632: 👍

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
malish8632: removed

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
malish8632: i've consolidated it

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
malish8632: I've updated this to use AtomicReference instead of an object itself.

agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
```diff
@@ -89,11 +89,15 @@ public void testConfMixedPluginsEnabled() {
     public void testConfigFromEnvVars() {
         environmentVariables.set("FFWD_TTL", "100");
         environmentVariables.set("FFWD_OUTPUT_RATELIMIT", "1000");
+        environmentVariables.set("FFWD_OUTPUT_CARDINALITYLIMIT", "10000");
+        environmentVariables.set("FFWD_OUTPUT_HYPERLOGLOGSWAPPERIODMS", "3000000");
```
dmichel1: i'm having a hard time reading `HYPERLOGLOGSWAPPERIODMS` can that be made more readable?

dmichel1 (approved):
nice - one more nit about naming then lgtm

agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
malish8632: done

PR #193, opened 2020-05-21
Convert batch to list of metrics and use cardinality of individual metric

malish8632:
This fix should normalize how we treat metrics for cardinality computation regardless how they are sent as batch or individual metric.
Also introducing BatchMetricConverter to be used instead of FakeBatchablePluginSinkBase as name is misleading.


hexedpackets (approved):
LGTM after fixing the checkstyle violations

PR #195, opened 2020-06-09
Introducing high frequency detector

### LINKED ISSUE #192: Detect when time between data points of the same series is too small
Detect when time between data points of the same series is too small and implement a limit on this

malish8632:
This change introduces high frequency detector that detects metrics sent with small delta time between data points. Initially it would only report current state of detection in **sendMetric** path avoiding any **sendBatch** methods.
Dropping metrics could be turned on by configuring **output** module.

This change also tries to simplify dependency tree of all modules.
New tree:
dependency-graph

Old tree:
dependency-graph-current

Closes #192


sjoeboo (approved):
LGTM

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public class HighFrequencyDetector {
+
+  /** Allow to drop high frequency metrics. */
+  @Inject
+  @Named("dropHighFrequencyMetric")
+  Boolean dropHighFrequencyMetric;
```
mattnworb: is it legit for these fields to be nullabe? just wondering if the primitive types could be used instead

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
```diff
@@ -0,0 +1,157 @@
...
+                        Collectors.toList(),
+                        list -> {
+                          int size = list.size();
+                          return IntStream.range(1, size)
+                              .map(
+                                  x ->
+                                      (int)
+                                          (list.get(size - x).getTime().getTime()
+                                              - list.get(size - x - 1).getTime().getTime()))
+                              .filter(d -> (d >= 0 && d < minFrequencyMillisAllowed))
```
mattnworb: perhaps extract this predicate and the map function above it (the part where two Metric instances have their timestamps subtracted) into private methods to lower the amount of indentation / operations here

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
```diff
@@ -0,0 +1,157 @@
...
+    this.highFrequencyTriggersTS = new AtomicLong(System.currentTimeMillis());
+    this.highFrequencyRecycleLock = new AtomicBoolean(false);
+  }
+
+  public List<Metric> detectHighFrequencyMetrics(final List<Metric> metrics) {
+    List<Metric> newList = new ArrayList<>();
+
+    // Groups metrics by metric hash code and finds timestamp deltas of ordered data points.
+    // {hashcode -> -1, hashcode1 -> 10}
+    Map<Integer, Integer> groupedMetrics =
```
mattnworb, line 97: looks like this Map is only used when `dropHighFrequencyMetric` is true, could you move that into the `if (dropHighFrequencyMetric) { ...` block to avoid computing this when it is not needed?

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
```diff
@@ -0,0 +1,157 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicReference;
+import java.util.stream.Collectors;
+import java.util.stream.IntStream;
+
+public class HighFrequencyDetector {
```
mattnworb, line 42: consider adding a brief description of what this class does, to help anyone unfamiliar with the codebase

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
```diff
@@ -0,0 +1,157 @@
...
+          c. check swap of high freq hashmap
+   */
+
+  @Inject
+  public HighFrequencyDetector() {
+    this.highFrequencyTriggersTS = new AtomicLong(System.currentTimeMillis());
+    this.highFrequencyRecycleLock = new AtomicBoolean(false);
+  }
+
+  public List<Metric> detectHighFrequencyMetrics(final List<Metric> metrics) {
```
mattnworb: Seems like this method is expected to sometimes filter the input and return a different List. It wasn't immediate obvious to me based on how it is used in BatchPluginSink that it might change the list, since the method name starts with `detect` - a quick javadoc comment about what the method does here could help.

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
```diff
@@ -0,0 +1,157 @@
...
+            });
+
+    statistics.reportHighFrequencyMetrics(highFrequencyMetrics.get().size());
+    swapHighFrequencyTriggersData();
+  }
+
+  /** This will reset high frequency triggers data */
+  private void swapHighFrequencyTriggersData() {
+    if (System.currentTimeMillis() - highFrequencyTriggersTS.get() > highFrequencyDataRecycleMS
+        && highFrequencyRecycleLock.compareAndExchange(false, true)) {
```
mattnworb: I'm a little confused on why an AtomicBoolean is used, could this just be `synchronized(some object) { .. }` or use a [`Lock`](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/locks/Lock.html)? seems like the highFrequencyRecycleLock's value is always reset back to `false` when this block ends - is the idea that if two threads call it at the same time, only one should enter (as opposed to a lock where both could grab the lock,  one after the other)? 

assuming a simple Lock doesn't fit the need, a comment here to explain the pattern could help

mattnworb (commented):
A few small comments and questions - also that module diagram looks like it could be a good addition to the docs in the repo itself

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
malish8632: Good point, I've changed it to use primitive as I default in case it's not set.

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
malish8632: Yep, moved.

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
malish8632: Actually I wanted it to be computed regardless of the flag to drop is true or not. So moved it a bit as we are interested in stats before we start dropping them.

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
malish8632: Done!

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
malish8632: Done! Also renamed it to shorter name as it's used on instance anyway and the name is self explanatory. I hope :)

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
malish8632: I've updated to use `synchronized(some object) { .. }` as it should be sufficient for our case.
The idea is to restart stats.

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
```diff
@@ -0,0 +1,167 @@
...
+  @Inject
+  private HighFrequencyDetectorStatistics statistics;
+
+  /** High frequency metrics counter. */
+  // TODO review to use ConcurrentHashMap vs AtomicReference
+  final AtomicReference<Map<Integer, Integer>> highFrequencyMetrics =
+      new AtomicReference<>(new HashMap<>());
+
+  final AtomicLong highFrequencyTriggersTS;
+  private final Object lock = new Object();
```
ao2017: Java now has an actual Lock. Have you considered using that

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
```diff
@@ -0,0 +1,167 @@
...
+  @Inject
+  @Named("highFrequencyDataRecycleMS")
+  long highFrequencyDataRecycleMS;
+
+  @Inject
+  private HighFrequencyDetectorStatistics statistics;
+
+  /** High frequency metrics counter. */
+  // TODO review to use ConcurrentHashMap vs AtomicReference
+  final AtomicReference<Map<Integer, Integer>> highFrequencyMetrics =
```
ao2017, line 68: Based on the usage of the variable, I think AtomicReference is better

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
```diff
@@ -0,0 +1,167 @@
...
+
+    return metrics;
+  }
+
+  private int computeTimeDelta(List<Metric> list) {
+    int size = list.size();
+    return IntStream.range(1, size)
+        .map(
+            x ->
+                (int)
```
ao2017, line 129: May be I am missing something :). I don't see the point of casting.  Even if the difference is small, it is good practice to avoid this type of casting.

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
```diff
@@ -0,0 +1,167 @@
...
+                  .compute(hashcode.getKey(), (key, val) -> (val == null) ? 1 : val + 1);
+            });
+
+    statistics.reportHighFrequencyMetrics(highFrequencyMetrics.get().size());
+    swapHighFrequencyTriggersData();
+  }
+
+  /** Resets high frequency triggers data hashmap */
+  private void swapHighFrequencyTriggersData() {
+    synchronized (lock) {
```
ao2017: Ha I see what you are  trying to do with with the lock. You probably don't need a lock why not synchronize the entire method  and  use methods that provide atomicity for multiple operations for example  compareAndSet ...

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
malish8632: Casting needed as the computations done on `long` values and if I remove casting I'm getting this

`Bad return type in lambda expression: long cannot be converted to int`

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
malish8632: I've decided to use `synchronized` block here.

mattnworb (approved):

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
malish8632: I think I can use `synchronized` method but not `compareAndSet` as I'm checking one atomic var and update two of them.

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
ao2017: I think this is because of the method return type which is  an int.

ao2017 (approved):

PR #196, opened 2020-06-12
Ignore small bursts of high freq metrics

malish8632:
This change will allow to ignore small bursts of high frequency metrics. 
Some of existing systems duplicate their metric production and that skews high frequency detector.

+ @ao2017

### DIFF
--- a/api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
+++ b/api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
@@ -43,6 +43,8 @@ import org.slf4j.Logger;
  */
 public class HighFrequencyDetector {
 
+  public static final int BURST_THRESHOLD = 5;
+
   /** Allow to drop high frequency metrics. */
   @Inject
   @Named("dropHighFrequencyMetric")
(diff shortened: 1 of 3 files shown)


sjoeboo (approved):
👍

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
```diff
@@ -123,15 +128,23 @@ public HighFrequencyDetector() {
...
                     (list.get(size - x).getTime().getTime()
                         - list.get(size - x - 1).getTime().getTime()))
         .filter(d -> (d >= 0 && d < minFrequencyMillisAllowed))
-        .min()
-        .orElse(-1);
+        .summaryStatistics();
+
+    int result = -1;
+
+    if (stats.getCount() > 5) {
```
ao2017: What is the rational of using 5 as the threshold?  It is better not to hardcode the threshold. 
Public static final int THRESHOLD = 5;  if(stats.getCount() > THRESHOLD)

malish8632: It's magic number :) - I'll change

ao2017: :), I am curious why 5

malish8632: It is based on preliminary testing across 65 canary instances.

ao2017 (approved):

PR #200, opened 2020-07-07
Add support for running FFWD with Corretto Java 11

smstone:
@lmuhlha @malish8632 @sjoeboo can you PTAL?

PR #203, opened 2020-07-14
Add java-11-amazon-corretto-jdk package dep

smstone:
@lmuhlha @malish8632 @sjoeboo @RochesterinNYC

https://github.com/spotify/ffwd/pull/200 only added support for running with Corretto if it was available. These changes allow to have Corretto, Oracle, or OpenJDK satisfy the Depends field.

### DIFF
--- a/agent/src/deb/control/control
+++ b/agent/src/deb/control/control
@@ -10,6 +10,6 @@ Build-Depends:
  maven (>= 3)
 Standards-Version: 3.9.1
 Architecture: all
-Depends: oracle-jvm-11 (>=11.0.2-1spotify1) | openjdk-11-jre-headless
+Depends: java-11-amazon-corretto-jdk-noalt (>=1:11.0.8.10-1spotify1) | oracle-jvm-11 (>=11.0.2-1spotify1) | openjdk-11-jre-headless
 Description: FastForward Java Agent
  FastForward is a metrics forwarding agent
--- a/agent/src/deb/bin/ffwd
+++ b/agent/src/deb/bin/ffwd
@@ -26,7 +26,7 @@ if [[ ! -f $JVMARGS ]]; then
 fi
 
 if [[ -z $JAVA ]]; then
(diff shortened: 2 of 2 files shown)


agent/src/deb/control/control
```diff
@@ -10,6 +10,6 @@ Build-Depends:
  maven (>= 3)
 Standards-Version: 3.9.1
 Architecture: all
-Depends: oracle-jvm-11 (>=11.0.2-1spotify1) | openjdk-11-jre-headless
+Depends: java-11-amazon-corretto-jdk (>=1:11.0.7.10-1) | oracle-jvm-11 (>=11.0.2-1spotify1) | openjdk-11-jre-headless
```
malish8632: if we add this dependency will it be installed during the build?

RochesterinNYC: The `|` operator for `Depends` is an or operator right? So this indicates that `java-11-amazon-corretto-jdk` should be installed to fulfill the dependency group (strung together by `|`s) but if it's unavailable, either of the other two can be installed?

RochesterinNYC (approved):

RochesterinNYC:
Noting also that this PR/change will allow VMs/runtimes that are using the version of `ffwd` that will be built by this PR to purge Java 11 Oracle. Before this PR/change, since `ffwd` is compatible with Java 11 Corretto but still declares Java 11 Oracle as a dependency, trying to purge or remove the Java 11 Oracle apt package would result in `ffwd` being uninstalled/removed as well.

agent/src/deb/control/control
lmuhlha: yep that's the intention

smstone:
@RochesterinNYC we can wait until https://github.com/corretto/corretto-11/issues/113 is addressed to move forward with this.

smstone:
@malish8632 @lmuhlha @RochesterinNYC We have the noalt package available to us now. Can you PTAL?

malish8632 (approved):
LGTM

lepistone (commented):
Looks great!

PR #206, opened 2020-07-24
Add tags to high freq metrics

malish8632:
This change should add more visibility into which metrics are high frequency.

### DIFF
--- a/agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
+++ b/agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
@@ -267,7 +267,7 @@ public class SemanticCoreStatistics implements CoreStatistics {
                 highFreqMetricsMap.put(gaugeMetric, (long) marked);
             }
 
-            private synchronized void registerGauge(MetricId gauge) {
+            private void registerGauge(MetricId gauge) {
                 if (highFreqMetricsMap.containsKey(gauge)) {
                     return;
                 }
--- a/agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
+++ b/agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
@@ -265,6 +265,22 @@ public class SemanticCoreStatistics implements CoreStatistics {
             public void reportHighFrequencyMetrics(int marked) {
                 highFreqMetrics.set(marked);
             }
+
+            public void reportHighFrequencyMetrics(int marked, String... tags){
+                /*
+                TODO finish up saving current num per metric to
+                    AtomicReference<Map<MetricId, Long>> or
+                    ConcurrentHashMap<MetricId, Long>()
+                 */
+
+                registry.register(m.tagged("what", "high-freq-metrics").tagged(tags), new Gauge<Integer>() {
+                    @Override
+                    public Integer getValue() {
+                        final int queueLength = marked;
+                        return queueLength;
+                    }
+                });
+            }
         };
     }
 }
--- a/agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
+++ b/agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
@@ -259,17 +259,18 @@ public class SemanticCoreStatistics implements CoreStatistics {
             }
 
             @Override
-            public void reportHighFrequencyMetrics(int marked, String... tags){
+            public void reportHighFrequencyMetrics(int marked, String... tags) {
                 MetricId gaugeMetric = m.tagged("what", "high-freq-metrics").tagged(tags);
 
                 registerGauge(gaugeMetric);
 
-                highFreqMetricsMap.put(gaugeMetric, (long)marked);
+                highFreqMetricsMap.put(gaugeMetric, (long) marked);
             }
 
-            private synchronized void registerGauge(MetricId gauge){
-                if(highFreqMetricsMap.containsKey(gauge))
+            private synchronized void registerGauge(MetricId gauge) {
+                if (highFreqMetricsMap.containsKey(gauge)) {
                     return;
+                }
                 registry.register(gauge, new Gauge<Long>() {
                     @Override
                     public Long getValue() {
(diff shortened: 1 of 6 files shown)


sjoeboo (approved):

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
```diff
@@ -100,13 +101,13 @@ public HighFrequencyDetector() {
     List<Metric> newList = new ArrayList<>();
 
     // Groups metrics by metric hash code and finds times deltas of ordered data points.
```
smstone: This comment needs updating since this switches to using Function.identity()

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
```diff
@@ -100,13 +101,13 @@ public HighFrequencyDetector() {
...
-    // {hashcode -> -1, hashcode1 -> 10}
-    Map<Integer, Integer> groupedMetrics =
+    // {metric1 -> -1, metric2 -> 10}
+    Map<Metric, Integer> groupedMetrics =
         metrics.stream()
             .sorted(Comparator.comparing(Metric::getTime))
             .collect(
                 Collectors.groupingBy(
-                    Metric::hashCode,
+                    Function.identity(),
```
smstone, line 110: What is the reasoning for switching to `Function.identity()` over metric hash code?

agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
```diff
@@ -246,24 +247,36 @@ public void reportMetricsDroppedByFilter(final int dropped) {
...
-                highFreqMetrics.set(marked);
+            public void reportHighFrequencyMetrics(int marked, String... tags) {
+                MetricId gaugeMetric = m.tagged("what", "high-freq-metrics").tagged(tags);
+
+                registerGauge(gaugeMetric);
+
+                highFreqMetricsMap.put(gaugeMetric, (long) marked);
+            }
+
+            private synchronized void registerGauge(MetricId gauge) {
```
ao2017: synchronizing a read access to a concurrent concurrent hashMap is off it defies the point of using it.

agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
```diff
@@ -246,24 +247,36 @@ public void reportMetricsDroppedByFilter(final int dropped) {
...
+                if (highFreqMetricsMap.containsKey(gauge)) {
```
ao2017, line 271: Is this check necessary ? If you want to avoid null value below  return a zero or optional .

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
ao2017: My guess is, because we are now using metric.

ao2017 (approved):

api/src/main/java/com/spotify/ffwd/util/HighFrequencyDetector.java
malish8632: correct

agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
malish8632: @ao2017 @smstone  I'm using the synchronized access here as I'd like to make sure the registration of a gauge won't produce an exception. If there race condition it's possible `highFreqMetricsMap.containsKey` returns false but by the time it tries to register new Gauge other thread already done it. My understanding of synchronized method/block that it won't allow this race condition to happen.

agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
malish8632: The registration of new Gauge actually throws an exception if same exactly Gauge is trying to be registered.
As we don't know all possible permutations of different tags of "whats" and "keys" by using synchronized method I'm making sure I register new Gauge only once.

agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
ao2017: Check register.register implementation  you don't need synchronized. 
 I also  prefer the option of handling exception instead of  adding synchronized with a check.

Synchronized only provides serialized access to  this method any thread can change the register outside of  this method. So in this case if you rely on synchronized for thread safety it can easily be broken.

malish8632: I've added try/catch to handle race condition.

smstone (approved):

agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
```diff
@@ -246,24 +247,40 @@ public void reportMetricsDroppedByFilter(final int dropped) {
...
+            public void reportHighFrequencyMetrics(int marked, String... tags) {
+                MetricId gaugeMetric = m.tagged("what", "high-freq-metrics").tagged(tags);
+
+                registerGauge(gaugeMetric);
+
+                highFreqMetricsMap.put(gaugeMetric, (long) marked);
+            }
+
+            private void registerGauge(MetricId gauge) {
+                if (highFreqMetricsMap.containsKey(gauge)) {
```
ao2017, line 271: You can remove this now.  The entire logic is now atomic. You get an exception if you try to insert again.  
am I missing anything ?

ao2017 (approved):

agent/src/main/java/com/spotify/ffwd/statistics/SemanticCoreStatistics.java
malish8632: The register method should be called only in case this is new permutation of tags, otherwise every time it will throw the exception and I'd like to avoid it.

PR #212, opened 2020-08-24
Replace System Rules with System Lambda

stefanbirkner:
System Lambda is more specific. It only wraps the part of the code that
reads the environment variables.


lmuhlha (approved):
thanks!

PR #213, opened 2020-09-08
[WIP] Add  Json Metric with distribution support

ao2017:
This is a WIP. This PR is intended for today's discussion.
 The metric version, v1  is not final. 
 I have a preference for keeping everything in this project but I am  opened to any suggestions.

### DIFF
--- a/api/src/main/java/com/spotify/ffwd/model/v1/Metric.java
+++ b/api/src/main/java/com/spotify/ffwd/model/v1/Metric.java
@@ -1,0 +1,87 @@
+/*-
+ * -\-\-
+ * FastForward API
+ * --
+ * Copyright (C) 2016 - 2018 Spotify AB
+ * --
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ * 
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ * 
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


api/src/main/java/com/spotify/ffwd/model/Batch.java
```diff
@@ -29,6 +29,7 @@
 import java.util.Optional;
 import lombok.Data;
 import lombok.EqualsAndHashCode;
+import com.spotify.ffwd.v1.Value;
```
malish8632, line 32: I think this needs to be removed

api/src/main/java/com/spotify/ffwd/model/v1/Metric.java
```diff
@@ -0,0 +1,87 @@
...
+import com.spotify.ffwd.v1.Value;
+import java.util.Date;
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+/**
+ * TODO: rename to 'Point' to follow along with internal guidelines.
```
malish8632, line 40: I guess this could be removed

api/src/main/java/com/spotify/ffwd/model/v1/Metric.java
```diff
@@ -0,0 +1,87 @@
...
+ */
+@Data
+@EqualsAndHashCode(of = {"key", "riemannTags", "tags"})
```
malish8632, line 43: Don't we need to use resource tags here? @lmuhlha 
I think riemannTags should be removed as we are not using those anymore.

PR #215, opened 2020-09-14
[WIP]Add a New Metric Data Model and Update SerializerProto100

ao2017:
**Background:**
Please refer to this [issue.](https://github.com/spotify/ffwd/issues/214l)

**Requirements:**
FFWD agent supports 2 protocols TCP and UPD. UDP metrics are serialized using Google Proto 
and TCP metrics are mostly handles through spotify100 serializer. The requirements for distribution 
support is to maintain the current core functionality. 
1. Ensure  FFWD agent new version  can handle both the new and old metrics. 
2. Ensure that  metrics from new and old version of FFW agent  can
    be handled upstream using the same pubsub topic. 

**What's New** 
1. New JSON metric data model 
2. Update protoSerializer100
3. Update various interface to support the new metric format.


core/src/main/proto/spotify_100.proto
```diff
@@ -20,6 +20,9 @@ message Metric {
 
     // resource "tags" associated to metric.
     map<string, string> resource = 5;
+
+    // Point value
+     Point point =  10;
```
malish8632: I don't think Point is relevant here. This `Metric` has new type of value. The `time` you have in `Point` already available in `Metric`.
I think you should use same approach as [here](https://github.com/spotify/ffwd-client-java/blob/master/ffwd-client/src/main/proto/protocol1.proto)

api/src/main/java/com/spotify/ffwd/model/v2/Metric.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.Map;
+import java.util.Set;
+import java.util.TreeMap;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+
+
+@Data
+@EqualsAndHashCode(of = {"key", "tags"})
+public class Metric implements Metrics {
```
malish8632, line 36: I'm hesitant to go ahead with this change. Why can't we use original `Metric` and `Batch` objects and use new `Value` type instead of `double`? The `Metric` and `Batch` objects are internal to FFWD and I don't believe we use them anywhere. This then will introduce only Value object and change the Metric and Batch as they are.

core/src/main/proto/spotify_100.proto
malish8632: Also why field number 10? I think the next number should be 6 if we introducing new field to proto.

malish8632 (changes requested):
I think we should not introduce new objects/versions if we could avoid that.
Current ways to receive metrics in FFWD:
1. UDP/Proto - adding ability to decode the message [v1](https://github.com/spotify/ffwd-client-java/blob/master/ffwd-client/src/main/proto/protocol1.proto) or [v0](https://github.com/spotify/ffwd-client-java/blob/master/ffwd-client/src/main/proto/protocol0.proto) should be sufficient. And internally using same `Metric` and `Batch` with `Value` instead of `double` should work.

2. TCP/JSON - same here but with JSON. Internally using same `Metric` and `Batch` with `Value` instead of `double` should work.

Sending to PubSub with updated spotify_100.proto to support new field `Value` only should work.
I also believe that Heroic consumers should be updated first (after rigorous testing) with this new schema so any new messages published to PubSub will be handled correctly.

In regards of the cadence I think this change shouldn't be deployed first - we should have ability to upgrade FFWD and with this change merged we won't be able to do so until we test all pipeline components. 
Let's come up with plan how we could test these changes without them merged into master. We should be 100% sure these changes are not adding anything funky to metrics transportation.

api/src/main/java/com/spotify/ffwd/model/v2/Metric.java
ao2017: Technically, It is not internal. Http Reporter and non java-clients  must conform to this format, right ? 
 I think this is risk free unless there is a parity between instrumentation side and ffwd agent.

core/src/main/proto/spotify_100.proto
ao2017: Dropwizard metrics are still sending value in double format. We are using the new definition for distribution only. So I cannot just change double into object.  
Because it is a opt- in feature.

api/src/main/java/com/spotify/ffwd/model/v2/Metric.java
malish8632: This Metric object is used internally only.

core/src/main/proto/spotify_100.proto
malish8632: Double value will be sent by old FFWD users and Value value will be sent by new FFWD users. What Point object represents what is not described already? Metric is bunch of tags and value. Old value is double, new value is Value that's what should be add as in here https://github.com/spotify/ffwd-client-java/blob/master/ffwd-client/src/main/proto/protocol1.proto

ao2017: Let setup a meeting and chat about this offline.  Unless we want to ask Heroic users to send new metric we cannot change this.

api/src/main/java/com/spotify/ffwd/model/v2/Metric.java
ao2017: Let's chat offline about this. Metric definition can be confusing. Http client is using this format even if the object is not shared.

ao2017:
> I think we should not introduce new objects/versions if we could avoid that.
> Current ways to receive metrics in FFWD:
> 
> 1. UDP/Proto - adding ability to decode the message [v1](https://github.com/spotify/ffwd-client-java/blob/master/ffwd-client/src/main/proto/protocol1.proto) or [v0](https://github.com/spotify/ffwd-client-java/blob/master/ffwd-client/src/main/proto/protocol0.proto) should be sufficient. And internally using same `Metric` and `Batch` with `Value` instead of `double` should work.
> 2. TCP/JSON - same here but with JSON. Internally using same `Metric` and `Batch` with `Value` instead of `double` should work.
> 
> Sending to PubSub with updated spotify_100.proto to support new field `Value` only should work.
> I also believe that Heroic consumers should be updated first (after rigorous testing) with this new schema so any new messages published to PubSub will be handled correctly.
> 
> In regards of the cadence I think this change shouldn't be deployed first - we should have ability to upgrade FFWD and with this change merged we won't be able to do so until we test all pipeline components.
> Let's come up with plan how we could test these changes without them merged into master. We should be 100% sure these changes are not adding anything funky to metrics transportation.

@malish8632  as discussed offline in last week meeting  and yesterday morning, we need to introduce second version of  the metrics to ensure compatibility. This is a common practice if you have an API. We will support both new and old metrics. As I stated during our monthly meeting on distribution support in heroic, it is an opt-in feature. Heroic users don't have to update any code unless they want to use distribution.  

Regarding deployment,  my goal is to introduce this feature  gradually because of its scope.  If anyone try to send a new version of the metric FFWD agent will drop it for now.  We will start sending distribution metric to heroic when heroic  will be able to handle it. 

For testing, we will use well known best practices: good unit test coverage and integration testing in staging env. So far, I  added  distribution to 2 projects.  In both cases, I  improved  the code coverage.  We are going to use the same approach for FFWD.

I hope that this change request is settled.

core/src/main/proto/spotify_100.proto
ao2017: As discussed offline, we cannot just replace the value format.   Also we cannot use oneof in this case. Please look at this as 2 versions in one file.  If you don't like the idea of sending new version  value wrap into a Point object. I will remove it. 
I think this conversation could have been avoided if we have a single global definition of metric the current state can easily lead to confusion and bugs.

malish8632: Hey @ao2017 it's not that I don't like this solution - I think that Point is useless and we could describe what we want with `Value newValue`. Let's remove Point as we agreed offline. Thanks

core/src/main/proto/spotify_100.proto
```diff
@@ -21,8 +21,8 @@ message Metric {
     // resource "tags" associated to metric.
     map<string, string> resource = 5;
 
-    // Point value
-     Point point =  10;
+    // Distribution type value
+     Value distribution_type_value = 6;
```
malish8632, line 25: I think this name a bit confusing - this should be something `value_obj` as it could support either double or distribution.

malish8632 (approved):
Minor change of name and could you also update PR description with a note that this PR won't change current implementation as discussed offline.
Thanks.

core/src/main/proto/spotify_100.proto
ao2017: value_obj  :). NP, I will change with the next PR.

PR #221, opened 2020-10-13
Add distribution support to ffwd agent modules

ao2017:
Heroic histogram data is currently computed locally.
It is practically impossible to aggregate percentile.
We are adding distribution to heroic to address that issue.
Distribution will create data sketch that will be used upstream to compute percentile on the entire data distribution.

This PR affects many files but some of the change simply change the metric type in the import statement.
The changes that are consequential are described below and in the attached document.

 **Metric Data Model**
 We changed the data point value type to [Value](url) and eliminate unused fields. 
     This change will have no performance implication unless  users start sending distribution metric.
     
 **Serialization:** 
 Even though ffwd will continue to process old version of metric, it will only forward metric into new version only.
 There is a PR pending that will handle  the new metric format in Heroic consumer. 
 We don't expect any performance implication related to this change since input metric conversion is not new. 

**ProtoBuf Module**
 Has been updated and tested to support distribution

 **Http Module**
Has been updated to handle both new and old metric.  
Dev for distribution support in http client is pending.

Related Documentation : 
https://docs.google.com/document/d/1OsRTPpU9AUNbyV-LPBG4TQFtPIcyISdrFuw3AREIy4M/edit?usp=sharing


core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -88,7 +86,7 @@
 
     @Inject
     @Named("riemannTags")
-    private Set<String> riemannTags;
+    private Set<String> riemannTags;  //TODO IS THIS NEEDED ??
```
lmuhlha: we got rid of riemann so it shouldnt be needed anymore

core/src/test/java/com/spotify/ffwd/serializer/TestSpotify100ProtoSerializer.java
```diff
@@ -176,57 +144,13 @@ public void testSerializedMetricWithDistributionValue() throws Exception {
   }
 
 
-  @Test
```
lmuhlha, line 179: should the removed tests in here be replaced or are they covered by something else now?

modules/http/src/main/java/com/spotify/ffwd/http/HttpDecoder.java
```diff
@@ -61,7 +64,7 @@ protected void decode(ChannelHandlerContext ctx, FullHttpRequest in, List<Object
                 }
 
                 throw new HttpException(HttpResponseStatus.METHOD_NOT_ALLOWED);
-            case "/v1/batch":
+            case "/v1/batch": case "/v2/batch":
```
lmuhlha, line 67: does this syntax just mean in either case do the following? I thought the v1/batch endpoint was removed above 🤔  `private static final String V1_BATCH_ENDPOINT = "v1/batch";`

ao2017: Yes it does, Regarding  "v1/batch",   ffwd-http-client will continue to send metric using that uri => https://github.com/spotify/ffwd-http-client/blob/master/src/main/java/com/spotify/ffwd/http/RawHttpClient.java#L39

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
ao2017: Alright, I will remove it then

modules/http/src/main/java/com/spotify/ffwd/http/HttpDecoder.java
lmuhlha: 👍  thanks for clarifying

core/src/main/proto/spotify_100.proto
```diff
@@ -12,23 +12,18 @@ message Metric {
     // time in ms when metric was generated.
     int64 time = 2;
 
-    // value of metric.
```
malish8632: Based on the [docs](https://developers.google.com/protocol-buffers/docs/proto#reserved) we shouldn't remove the fields but mark them **reserved**

modules/http/src/main/java/com/spotify/ffwd/http/HttpClient.java
```diff
@@ -36,7 +36,7 @@
 import okhttp3.Response;
 
 public class HttpClient {
-    private static final String V1_BATCH_ENDPOINT = "v1/batch";
+    private static final String V2_BATCH_ENDPOINT = "v2/batch";
```
malish8632, line 39: How are we going to switch all clients of metricsapi that currently sending to "v1/batch" to start using "v2/batch"? I think it will break those clients. I think we should support both end points and in "v1/batch" one convert submitted metrics to new v2 metric.

modules/json/src/main/java/com/spotify/ffwd/json/JsonObjectMapperDecoder.java
```diff
@@ -105,27 +105,24 @@ private Object decode0(ByteBuf in, List<Object> out) throws IOException {
...
+        return new Metric(key, Value.DoubleValue.create(value), time, tags, resource);
     }
 
-    private Date decodeTime(JsonNode tree, String name) {
+    private long decodeTime(JsonNode tree, String name) {
         final JsonNode n = tree.get(name);
 
         if (n == null) {
-            return null;
+            return 0;
```
malish8632, line 122: Did you test that this would create metric with current timestamp? I believe the old code by setting it to null insures it will be set to current time downstream.

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
```diff
@@ -32,8 +32,8 @@
 import com.google.pubsub.v1.ProjectTopicName;
 import com.google.pubsub.v1.PubsubMessage;
 import com.spotify.ffwd.cache.WriteCache;
-import com.spotify.ffwd.model.Batch;
-import com.spotify.ffwd.model.Metric;
+import com.spotify.ffwd.model.v2.Batch;
+import com.spotify.ffwd.model.v2.Metric;
```
malish8632, line 36: This change is a bit sensitive. I'm curious how we could verify that Heroic consumer will be able to consume new metric serialized with updated Spotify100ProtoSerializer.java that uses v2 metric?

malish8632:
There are couple of points I've raised in regards of publishing metrics to PubSub with new serializer - not sure if the integration with Heroic consumer was ever tested. I think we should make sure it passes that test.
Also there were multiple new empty lines added across several files - could you remove them?

modules/http/src/main/java/com/spotify/ffwd/http/HttpClient.java
ao2017: There is no need to switch every client to v2/batch.  Let's look at  the "pipeline" 

httpclient ->httpDecoder -> InputChannel   ---->pubsub.   

Since httpDecoder can support both v1/batch and v2/batch there is no need to switch every client.
Essentially, we use the uri to determine the format of the metric in this case we have to use v2/batch otherwise it won't work.

modules/json/src/main/java/com/spotify/ffwd/json/JsonObjectMapperDecoder.java
ao2017: I changed the data format. It used to be Date in V2 it is long. So we now set to the current timestamp when the value is 0.

modules/http/src/main/java/com/spotify/ffwd/http/HttpClient.java
malish8632: I see. Thank you 👍

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
ao2017: I added decent test coverage in ffwd and  on herioc side. If you still have some concern, you can check locally. I can send you a code for a client that generate both new and old metric.

ao2017:
Which test are you talking about ?  Shouldn't we trust our build process.

I am sorry if I added extra lines.  Where do see the empty lines ? I will revert those changes.

Making  changes to a huge codebase with  many components require a lot of considerations. If we can remove style preference from that equation, I think that would be a win.    You may want to  update the checkstyle to reflect your preference. We will catch these issues before PR submission.

malish8632:
If it's possible to create ffwd docker image manually and test on staging to see that Heroic consumers are capable to consume new messages before merging it, that would be great.
I understand the size of the PR. If you could just glance over this PR here in github you'll notice some additional lines.

malish8632 (approved):
Would be great to test on staging before merging. Otherwise 👍

lmuhlha (approved):

ao2017:
I closed the pr by mistake

PR #223, opened 2020-10-22
Add tag refreshing functionality

davidxia:
PR is WIP. Please make any edits or suggestions.

`/Users/dxia/Downloads/work-notes/labels.yaml` looks like

```
cluster="test-cluster1"
rack="rack-22"
tugboat.spotify.com/stage="canary"
zone="us-est-coast"
```

Running this locally with the automatically enabled DebugPluginSink and the command

```
cat metric.json

{
  "points": [
    {
      "key": "test_key",
       "tags": {
         "what": "error-rate"
       },
       "resource": {},
       "value": 1234.0,
       "timestamp":11111
    }
  ]
}

cat metric.json | http POST http://localhost:8080/v1/batch 'Content-Type:application/json'
```

I get

```
19:32:35.447 [main] INFO  com.spotify.ffwd.AgentCore - Started!

19:33:57.723 [ffwd-scheduler-0] INFO  com.spotify.ffwd.debug.DebugPluginSink -
B#0: Batch(commonTags={cluster="test-cluster1", rack="rack-22",
tugboat.spotify.com/stage="canary", zone="us-est-coast", foo=bar, env=prod},
commonResource={}, points=[Batch.Point(key=test_key, tags={what=error-rate},
resource={}, value=1234.0, timestamp=11111)])
```

Then I change the contents of `labels.yaml` to have
`tugboat.spotify.com/stage="production"` and run the same command and get

```
19:34:40.382 [ffwd-scheduler-1] INFO  com.spotify.ffwd.debug.DebugPluginSink -
B#1: Batch(commonTags={cluster="test-cluster1", rack="rack-22",
tugboat.spotify.com/stage="production", zone="us-est-coast", foo=bar,
env=prod}, commonResource={}, points=[Batch.Point(key=test_key,
tags={what=error-rate}, resource={}, value=1234.0, timestamp=11111)])
```

This shows the tag is being refreshed.

### DIFF
--- a/agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
+++ b/agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
@@ -21,6 +21,10 @@
 package com.spotify.ffwd;
 
 import static com.github.stefanbirkner.systemlambda.SystemLambda.withEnvironmentVariable;
+import static java.util.concurrent.TimeUnit.SECONDS;
+import static org.awaitility.Awaitility.await;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.hasEntry;
 import static org.junit.Assert.assertEquals;
 
 import com.google.common.collect.ImmutableList;
@@ -33,6 +37,7 @@ import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
 import org.junit.Test;
@@ -113,6 +118,34 @@ public class FfwdConfigurationTest {
         assertEquals("jimjam", outputManager.getHost());
         assertEquals(Long.valueOf(10000), outputManager.getCardinalityLimit());
         assertEquals(Long.valueOf(555555), outputManager.getHLLPRefreshPeriodLimit());
+    }
+
+    @Test
+    public void testDynamicTagsFile() throws Exception {
+        Path configPath = resource("basic-settings-with-dynamic-tags-file.yaml");
+        CoreOutputManager outputManager = withEnvironmentVariable("FFWD_TAG_foo", "bar")
+            .and("FFWD_TAG_other", "constant")
+            .and("IS_FFWD_CONFIGURATION_TEST", "1")
+            .execute(() -> getOutputManager(configPath));
+
+        // At first tags should have the foo=bar entry set by the env var above.
+        assertThat(outputManager.getTags(), hasEntry("foo", "bar"));
+        // Then the tag value should be updated from bar to qux.
+        await().atMost(5, SECONDS).until(
+            () -> {
+                final Map<String, String> tags = outputManager.getTags();
+                return Objects.equals(tags.get("foo"), "qux")
+                       && Objects.equals(tags.get("other"), "constant")
+                       && tags.get("not-specified") == null;
+            });
+    }
+
+    @Test
+    public void testDynamicTagsFileDoesNotExist() throws Exception {
+        Path configPath = resource("basic-settings-with-non-existent-dynamic-tags-file.yaml");
+        CoreOutputManager outputManager = withEnvironmentVariable("FFWD_TAG_foo", "bar")
+            .execute(() -> getOutputManager(configPath));
+        assertThat(outputManager.getTags(), hasEntry("foo", "bar"));
     }
 
     private CoreOutputManager getOutputManager(final Path configPath) {
--- a/core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
+++ b/core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
@@ -20,9 +20,11 @@
 
 package com.spotify.ffwd.output;
 
-import net.agkn.hll.HLL;
+import com.google.common.annotations.VisibleForTesting;
+import com.google.common.base.Strings;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
+import com.google.common.util.concurrent.ThreadFactoryBuilder;
 import com.google.inject.Inject;
 import com.google.inject.name.Named;
 import com.spotify.ffwd.debug.DebugServer;
@@ -33,6 +35,10 @@ import com.spotify.ffwd.statistics.OutputManagerStatistics;
 import com.spotify.ffwd.util.BatchMetricConverter;
 import eu.toolchain.async.AsyncFramework;
 import eu.toolchain.async.AsyncFuture;
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.util.AbstractMap.SimpleImmutableEntry;
 import java.util.Arrays;
 import java.util.Collections;
@@ -42,13 +48,16 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.Executors;
+import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 import javax.annotation.Nullable;
-
+import net.agkn.hll.HLL;
 import net.agkn.hll.HLLType;
 import org.isomorphism.util.TokenBucket;
 import org.isomorphism.util.TokenBuckets;
@@ -79,6 +88,11 @@ public class CoreOutputManager implements OutputManager {
     @Inject
     @Named("tags")
     private Map<String, String> tags;
+
+    @VisibleForTesting
+    public Map<String, String> getTags() {
+        return tags;
+    }
 
     @Inject
     @Named("tagsToResource")
@@ -125,10 +139,14 @@ public class CoreOutputManager implements OutputManager {
     @Inject
     private Filter filter;
 
+    @Inject
+    private String dynamicTagsFile;
+
     private AtomicReference<HLL> hyperLog;
     private AtomicLong hyperLogSwapTS;
     private AtomicBoolean hyperLogSwapLock;
     private Long hyperLogLogPlusSwapPeriodMS;
+    private ScheduledExecutorService tagRefreshingThread = null;
 
     public final Long getRateLimit() {
         if (rateLimiter == null) {
@@ -148,7 +166,8 @@ public class CoreOutputManager implements OutputManager {
     @Inject
     CoreOutputManager(@Named("rateLimit") @Nullable Integer rateLimit,
         @Named("cardinalityLimit") @Nullable Long cardinalityLimit,
-        @Named("hyperLogLogPlusSwapPeriodMS") @Nullable Long hyperLogLogPlusSwapPeriodMS) {
+        @Named("hyperLogLogPlusSwapPeriodMS") @Nullable Long hyperLogLogPlusSwapPeriodMS,
+        @Named("dynamicTagsFile") @Nullable String dynamicTagsFile) {
 
         if (rateLimit != null && rateLimit > 0) {
             // Create a rate limiter with a configurable QPS, and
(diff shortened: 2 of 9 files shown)


agent/ffwd.yaml
```diff
@@ -26,6 +26,7 @@ input:
         port: 8080
 
 output:
+  dynamicTagsFile: /Users/dxia/Downloads/work-notes/labels.yaml
```
ao2017: @davidxia  This doesn't look right.  How are you planning to provide the file path  in prod?

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -181,6 +194,27 @@ public final Long getHLLPRefreshPeriodLimit() {
         } else {
             this.cardinalityLimit = null;
         }
+
+        if (!Strings.isNullOrEmpty(dynamicTagsFile)) {
+            final File file = new File(dynamicTagsFile);
+            if (file.exists() && file.canRead()) {
+                tagRefreshingThread = Executors.newSingleThreadScheduledExecutor(
+                    new ThreadFactoryBuilder().setNameFormat("tag-refreshing-thread").build());
```
ao2017, line 210: Is the used of an executor here justified ? It inside a constructor the tag reference is not available to any thread yet.

agent/ffwd.yaml
davidxia: Thanks for asking. I wanted some feedback here. This line is just placeholder for illustration. I'll remove later.

1. What's the best place for this setting?

I was thinking it would be optional in the `output` section of this file. In our case we can add it to ffwd-java-shim here https://ghe.spotify.net/monitoring/ffwd-java-shim/blob/master/ffwd.yaml.

2. Is `dynamicTagsFile` OK? Feel free to suggest other names.

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
davidxia: Oh, can you explain why the tag reference isn't available? It seemed like it was? What did you have in mind instead?

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -181,6 +194,27 @@ public final Long getHLLPRefreshPeriodLimit() {
...
+    }
+
+    private void refreshTagsFromFile(final Path path) {
+        try (final Stream<String> lines = Files.lines(path)) {
+            lines.forEach(label -> {
+                final String[] labelParts = label.split("=", 2);
+                tags.put(labelParts[0], labelParts[1]);
+            });
+        } catch (IOException e) {
+            throw new RuntimeException(e);
```
davidxia: What should happen if the file isn't readable?

lmuhlha: This is where I was thinking maybe we fall back to the previous functionality or something along those lines

davidxia: So just log a warning and ignore?

davidxia: done

agent/ffwd.yaml
davidxia: Discussed with @lmuhlha and @adsail who said this works.

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -181,6 +200,34 @@ public final Long getHLLPRefreshPeriodLimit() {
         } else {
             this.cardinalityLimit = null;
         }
+
+        if (!Strings.isNullOrEmpty(dynamicTagsFile)) {
+            final File file = new File(dynamicTagsFile);
+            if (file.exists() && file.canRead()) {
+                final int initialDelay =
+                    System.getenv().containsKey("IS_FFWD_CONFIGURATION_TEST") ? 3 : 60;
```
davidxia, line 208: This is ugly test logic, but I couldn't think of a less ugly way to signal to this code here to set a short delay for the executor.

agent/src/test/java/com/spotify/ffwd/FfwdConfigurationTest.java
```diff
@@ -115,6 +120,34 @@ public void testMergeOrder() throws Exception {
...
+        Path configPath = resource("basic-settings-with-dynamic-tags-file.yaml");
+        CoreOutputManager outputManager = withEnvironmentVariable("FFWD_TAG_foo", "bar")
+            .and("FFWD_TAG_other", "constant")
+            .and("IS_FFWD_CONFIGURATION_TEST", "1")
+            .execute(() -> getOutputManager(configPath));
+
+        // At first tags should have the foo=bar entry set by the env var above.
+        assertThat(outputManager.getTags(), hasEntry("foo", "bar"));
+        // Then the tag value should be updated from bar to qux.
+        await().atMost(5, SECONDS).until(
```
davidxia, line 134: See https://github.com/awaitility/awaitility/wiki/Usage#simple

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -181,6 +200,34 @@ public final Long getHLLPRefreshPeriodLimit() {
...
+            } else {
+                log.warn("Cannot read file at {}. Ignoring it.", dynamicTagsFile);
+            }
+        }
+    }
+
+    private void refreshTagsFromFile(final Path path) {
+        try (final Stream<String> lines = Files.lines(path)) {
+            lines.forEach(label -> {
+                final String[] labelParts = label.split("=", 2);
```
davidxia, line 222: Assumption here is that tag key and values are separated by `=`.

davidxia:
Please link to any docs I should update. I saw the docs dir, but it didn't seem to have much.

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
malish8632: I'm curious if instead of this approach we should try to use yaml style configuration as we have in standard Apollo?
https://github.com/lightbend/config
Based on examples the labels.yaml file is yaml and I'm curious if we going to have more control of the labels.

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -181,6 +200,34 @@ public final Long getHLLPRefreshPeriodLimit() {
...
+                // Only update tag values if the tag keys are explicitly specified in normal configs
+                if (tags.get(labelParts[0]) != null) {
+                    tags.put(labelParts[0], labelParts[1]);
+                }
+            });
+        } catch (IOException e) {
+            // Ignore
```
malish8632, line 230: I was thinking about this case and I kinda think we should exit here with error - the main idea that this code will be executed only in case user indicated the tags are coming via labels.yaml file and if file either corrupted or doesn't exist the ffwd should exit as it does when configPath is not set or null. 
What you think @lmuhlha @hexedpackets?

core/src/test/java/com/spotify/ffwd/output/OutputManagerTest.java
```diff
@@ -139,6 +139,7 @@ protected void configure() {
                 bind(DebugServer.class).toInstance(debugServer);
                 bind(OutputManagerStatistics.class).toInstance(statistics);
                 bind(Filter.class).toInstance(filter);
+                bind(String.class).annotatedWith(Names.named("dynamicTagsFile")).toInstance("");
```
davidxia, line 142: Lmk if the empty string here is OK.

core/src/main/java/com/spotify/ffwd/output/OutputManagerModule.java
```diff
@@ -78,7 +79,8 @@ public OutputManagerModule(
         @JsonProperty("dropHighFrequencyMetric") @Nullable Boolean dropHighFrequencyMetric,
         @JsonProperty("minFrequencyMillisAllowed") @Nullable Integer minFrequencyMillisAllowed,
         @JsonProperty("minNumberOfTriggers") @Nullable Integer minNumberOfTriggers,
-        @JsonProperty("highFrequencyDataRecycleMS") @Nullable Long highFrequencyDataRecycleMS) {
+        @JsonProperty("highFrequencyDataRecycleMS") @Nullable Long highFrequencyDataRecycleMS,
+        @JsonProperty("dynamicTagsFile") @Nullable String dynamicTagsFile) {
```
davidxia, line 83: I made it nullable. Lmk if this is not OK.

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
davidxia: Which labels.yaml are you referring to?

The K8s downward API doesn't let you configure the format of the file it creates [1]. It'll always be like this.

```
key1=val1
key2=val2
```

1: https://kubernetes.io/docs/tasks/inject-data-application/downward-api-volume-expose-pod-information/#store-pod-fields

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
adsail: @malish8632 -- @lmuhlha and I spoke with @davidxia yesterday; plan to ignore is because the dynamic tags file will be exclusive to the new/ACA tags _for now_, leaving all pre-existing tags configured as they currently are. Moving forward, we can re-evaluate as to whether or not the file becomes the single source for tags, at which point we would handle this Exception differently.

davidxia: For tags that are known to be static for the lifetime of the process, I think it's fine to continue using env vars. So you can limit the use of this functionality to this use case: 1) you have tags that are known to change 2) and you can't restart the process with updated env vars.

davidxia:
Lmk how to move this PR along. I'm happy to walk through it or answer questions over hangouts.

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
malish8632: sounds good.

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
malish8632: Makes sense. Thanks

malish8632 (approved):
@davidxia Looks good to me. I'm curious how we will test this one - before the merge or after?

davidxia:
Good question. Before merging this, I can build the JAR locally, build the ffwd-java-shim Docker image locally. Run those on a test cluster to confirm it all works. Then we can merge this PR. Does that sound OK?

malish8632:
That would be great! Sorry for delay in responses.

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -181,6 +200,35 @@ public final Long getHLLPRefreshPeriodLimit() {
...
+            }
+        }
+    }
+
+    private void refreshTagsFromFile(final Path path) {
+        try (final Stream<String> lines = Files.lines(path)) {
+            lines.forEach(label -> {
+                final String[] labelParts = label.split("=", 2);
+                // Only update tag values if the tag keys are explicitly specified in normal configs
+                if (tags.get(labelParts[0]) != null) {
```
davidxia, line 224: ugh, just realized this doesn't work right now. We set a ffwd common tag called `stage` that takes it's value from the K8s label `tugboat.spotify.com/stage`. So in the K8s downward API file, it'll be `tugboat.spotify.com/stage=[canary|production]`. And `tugboat.spotify.com/stage` != `stage` here so the tag won't be included.

Our options are to add prefix-stripping logic here so that we can specify which prefixes to remove like `tugboat.spotify.com/` 🤮  Or we modify tugboat to use the label `stage` instead of `tugboat.spotify.com/stage` (with a migration period where we have both labels for a while with the same values).

I kind of favor the latter but also realize `stage` is pretty generic. The intention of the long label was to namespace it so it's clear it's from tugboat.

Thoughts?

malish8632: Yeah, it does look ugly. Adding some sort of label mapping into FFWD configuration also sounds like too complicated.
I like the latter as well 👍

lmuhlha: +1 the latter would be preferable

davidxia: Cool we will do that. Are there any docs I should update before this is merged?

core/src/main/java/com/spotify/ffwd/output/CoreOutputManager.java
```diff
@@ -181,6 +200,35 @@ public final Long getHLLPRefreshPeriodLimit() {
...
+            } else {
+                log.warn("Cannot read file at {}. Ignoring it.", dynamicTagsFile);
+            }
+        }
+    }
+
+    private void refreshTagsFromFile(final Path path) {
+        try (final Stream<String> lines = Files.lines(path)) {
+            lines.forEach(label -> {
+                final String[] labelParts = label.split("=", 2);
```
davidxia, line 222: @malish8632 @lmuhlha @adsail any thoughts on adding some debug logs here saying what tags have been refreshed? Is the log level right now INFO? A log message here would happen every minute.

I can always add later if we need it.

davidxia:
If no objections, I'll merge EOD today?

PR #225, opened 2020-11-06
don't setDistributionTypeValue if the metric is not a distribution type

samfadrigalan:
The current way non-distribution metrics are being built conflicts with [this condition](https://github.com/spotify/heroic/blob/de3489066ce029137885014410aafd529c85ab2a/heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100Proto.java#L106) in the consumer. This PR makes sure to only set the metric [distribution_type_value](https://github.com/spotify/ffwd/blob/master/core/src/main/proto/spotify_100.proto#L25) if the metric is a distribution.


ao2017:
@samfadrigalan  as discussed offline, the "if statement" can handle every type of metrics.  
Since metrics generated by apollo are working fine. I think the problem is a data issue.

samfadrigalan:
not needed

PR #226, opened 2021-01-05
Fix http data conversion

samfadrigalan:
#### Notes
This is PR fixes ffwd http module deserialization errors when consuming the older metrics schema (w/o distributions).


malish8632 (approved):
+1

PR #229, opened 2021-01-16
Compile/build for java 8+

lmuhlha:


ao2017 (approved):
LGTM. It is unfortunate this was upgraded to 11 then downgraded :).

PR #232, opened 2021-01-21
Add new output plugin for sending OpenTelemetry metrics over gRPC.

### LINKED ISSUE #228: Add an OpenTelemetry output plugin
Protobuf file: https://github.com/open-telemetry/opentelemetry-proto/blob/master/opentelemetry/proto/metrics/v1/metrics.proto

This will likely eventually supersede the OpenCensus plugin that was [introducted for StackDriver metrics](https://github.com/spotify/ffwd/pull/185) as OpenTelemetry is replacing OpenCensus/OpenTracing.

hexedpackets:
Currently batches and distibutions are not supported. I plan to add them in later if this turns out to be useful.

Closes #228


modules/opentelemetry/src/main/java/com/spotify/ffwd/opentelemetry/OpenTelemetryPluginSink.java
```diff
@@ -0,0 +1,174 @@
...
+        try {
+            //noinspection ResultOfMethodCallIgnored
+            stub.export(request);
+        } catch (Exception e) {
+            log.error(e.getMessage());
+        }
+    }
+
+    @Override
+    public void sendBatch(Batch batch) {
```
ao2017, line 107: I wonder why batch is not supported ?

modules/opentelemetry/src/main/java/com/spotify/ffwd/opentelemetry/OpenTelemetryPluginSink.java
```diff
@@ -0,0 +1,174 @@
...
+            io.opentelemetry.proto.metrics.v1.Metric.newBuilder()
+                .setName(name)
+                .setDoubleGauge(gauge);
+
+        String unit = tags.get("unit");
+        if (unit != null) {
+            metricBuilder.setUnit(unit);
+        }
+
+        ExportMetricsServiceRequest request = ExportMetricsServiceRequest.newBuilder()
```
ao2017, line 90: How is this metric used upstream ?

modules/opentelemetry/src/main/java/com/spotify/ffwd/opentelemetry/OpenTelemetryPluginSink.java
```diff
@@ -0,0 +1,174 @@
...
+import java.util.stream.Collectors;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class OpenTelemetryPluginSink implements PluginSink {
+    private static final Logger log = LoggerFactory.getLogger(OpenTelemetryPluginSink.class);
+
+    @Inject private AsyncFramework async;
+    private String endpoint;
+    private Map<String, String> headers;
```
ao2017, line 53: Map is not thread safe. Unless each instance is confined to a single thread this could be a problem

modules/opentelemetry/src/main/java/com/spotify/ffwd/opentelemetry/OpenTelemetryPluginSink.java
```diff
@@ -0,0 +1,174 @@
...
+    private ManagedChannel channel;
+    private MetricsServiceGrpc.MetricsServiceBlockingStub stub;
```
ao2017, line 55: same as above if sub is not thread safe

modules/opentelemetry/src/main/java/com/spotify/ffwd/opentelemetry/OpenTelemetryPluginSink.java
hexedpackets: Its just not necessary right now, and the best way to map a batch to the OT metric definition isn't 100% clear so I'm kicking the can for the moment.

modules/opentelemetry/src/main/java/com/spotify/ffwd/opentelemetry/OpenTelemetryPluginSink.java
hexedpackets: Depends on the receiver. For Lightstep, the metrics are grouped on the name and then allow tag filtering/grouping under that.

modules/opentelemetry/src/main/java/com/spotify/ffwd/opentelemetry/OpenTelemetryPluginSink.java
hexedpackets: Its immutable, values are only set when initially parsing the config.

modules/opentelemetry/src/main/java/com/spotify/ffwd/opentelemetry/OpenTelemetryPluginSink.java
hexedpackets: Stubs are threadsafe, the base abstract class is annotated accordingly.

https://grpc.github.io/grpc-java/javadoc/io/grpc/stub/AbstractBlockingStub.html

ao2017 (approved):

PR #233, opened 2021-01-22
Refactor Batch.Point to Metric

hexedpackets:
Batch.Point and Metric have almost the exact same definition, the only significant differences were that Batch.Point has a JSON constructor and uses `timestamp` instead of `time`. I refactored to only use Metrics across ffwd.

Unrelated small change that I'm combining in this PR is to make the GeneratedPluginSource count configurable.


malish8632 (approved):
+1

PR #235, opened 2021-01-22
Revert Docker image to Java 11

hexedpackets:
All the Docker Hub builds are failing since the base image was switched to 8. AFAIK there's no reason to use 8 in Docker since it would be totally independent from any projects requiring Java 8 (@lmuhlha correct me if I'm wrong) so I'm switching it back to 11.


malish8632 (approved):
I don't think we need 8 in Docker either.

PR #246, opened 2021-02-25
Resize pubsub batches when they are above 10MB limit

lmuhlha:
When a batch is sent that is over the pubsub limit of 10MB we get 
`ERROR com.spotify.ffwd.pubsub[output:2] - Failed sending metrics io.grpc.StatusRuntimeException: INVALID_ARGUMENT: Request payload size exceeds the limit: 10485760 bytes.`
Which gets retried and causes a backup of metrics.

This will check the size of metrics being sent in bytes and reduce the size appropriately before sending to pubsub.

### DIFF
--- a/modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
+++ b/modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
@@ -26,6 +26,7 @@ import com.google.api.core.ApiFutures;
 import com.google.api.gax.rpc.NotFoundException;
 import com.google.cloud.pubsub.v1.Publisher;
 import com.google.common.cache.Cache;
+import com.google.common.collect.Lists;
 import com.google.common.util.concurrent.MoreExecutors;
 import com.google.inject.Inject;
 import com.google.protobuf.ByteString;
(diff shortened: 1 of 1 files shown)


modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
```diff
@@ -116,7 +121,18 @@ public void onSuccess(String messageId) {
 
     try {
       final ByteString m = ByteString.copyFrom(serializer.serialize(metrics, writeCache));
-      publishPubSub(m);
+
+      if (m.size() > MAX_BATCH_SIZE_BYTES) {
+        logger.debug("Above byte limit, resizing batch", metrics.size());
+        int times = (int)Math.ceil(m.size()/MAX_BATCH_SIZE_BYTES);
```
adsail, line 127: Edit: was thinking 

`times = m.size() % MAX_BATCH_SIZE_BYTES == 0 ?  m.size() / MAX_BATCH_SIZE_BYTES : m.size() / MAX_BATCH_SIZE_BYTES + 1`

adsail: And then could make the constant an int or long

malish8632 (approved):

PR #247, opened 2021-02-25
Use metrics size, not in bytes for pubsub limit fix

lmuhlha:


malish8632 (approved):

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
```diff
@@ -125,7 +125,7 @@ public void onSuccess(String messageId) {
       if (m.size() > MAX_BATCH_SIZE_BYTES) {
         logger.debug("Above byte limit, resizing batch");
         int times = (int)Math.ceil(m.size()/MAX_BATCH_SIZE_BYTES);
-        List<List<Metric>> collections = Lists.partition(new ArrayList<>(metrics), m.size()/times);
+        List<List<Metric>> collections = Lists.partition(new ArrayList<>(metrics), metrics.size()/times);
```
hexedpackets, line 128: I'm a little confused by this. Why don't you want to use byte size? And wouldn't this mess up the partitioning since both the limit and the `times` variable are using bytes?

hexedpackets: Actually I get why you want the list to be sized based on the number of metrics. But I still have the question about the limit and times being byte-based.

lmuhlha: Because that is breaking up the list of metrics so you need to divide the list of metrics size by the value of times. 
Before it was ending up as one collection and being sent the same as before because m.size()/times is huge compared to the size of the list of metrics. So you want to partition:
`Lists.partition(list_of_metrics, size_of_partitions_you_want_from_that_list);`

lmuhlha: image

lmuhlha: Before: 
Screen Shot 2021-02-25 at 5 13 06 PM

lmuhlha: The limit and times are byte based because the limit is 10MB, so you want to compare the input bytes to that and if it's bigger, figure out how many times you need to divide it to make batches that are < 10MB

hexedpackets: Makes sense now, thanks for explaining

hexedpackets (approved):

PR #249, opened 2021-02-26
[PubsubPluginSink] Drop requests greater than MAX_INPUT_METRICS

lmuhlha:


modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
```diff
@@ -119,22 +122,32 @@ public void onSuccess(String messageId) {
...
+              m.size());
+          int times = (int) Math.ceil(m.size() / MAX_BATCH_SIZE_BYTES);
+          List<List<Metric>> collections =
+              Lists.partition(new ArrayList<>(metrics), metrics.size() / times);
+          for (List<Metric> l : collections) {
+            try {
+              final ByteString mResize = ByteString.copyFrom(serializer.serialize(l, writeCache));
+              publishPubSub(mResize);
+            } catch (Exception e) {
+              logger.error("Failed to serialize batch of metrics: ", e);
```
samfadrigalan, line 139: should we throw an HttpException where we're catching these exceptions?

hexedpackets: I don't think this is in the path of HTTP requests, the input plugins are done once ffwd has buffered the metrics internally.

lmuhlha: I don't think I can throw an exception at that point, we'd need to figure out the size sooner somewhere upstream

hexedpackets (approved):

modules/pubsub/src/main/java/com/spotify/ffwd/pubsub/PubsubPluginSink.java
lmuhlha: lol hey william

